PR #28, opened 2024-03-28
Fix Iterable Dataset bug on downloading the whole subset of data

bernardhan33:
While working on the baseline performance numbers of Dataflux Iterable Dataset I found this critical bug that the previous implementation can ask each worker to download the whole dataset before applying `data_format_fn` to each item.

The bug is caused by using using the `[]` bracket to gather the results which break the generator pattern. Instead, we should use `()`. See more https://lerner.co.il/2018/06/08/python-parentheses-primer/.

The fix is tested by running the unit tests again and also rerun with the large dataset that exceeds the memory. Make sure that the memory utilization is correct. https://screenshot.googleplex.com/8g6A9qtHAiPGQ6r

As a retro:
1. The bug is introduced but didn't get caught by the unit test because in unit test we didn't test the logic of lazy generation of data. It is also likely that we can't test it because 1. the return of `dataflux_download_lazy` is mocked and 2. hard to inject into the code flow to verify that "yield from" is operated on a generator than a list.
2. For the performance tests, we should start to monitor memory usage.

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/dataflux_iterable_dataset.py
+++ b/dataflux_pytorch/dataflux_iterable_dataset.py
@@ -111,7 +111,7 @@ class DataFluxIterableDataset(data.IterableDataset):
         worker_info = data.get_worker_info()
         if worker_info is None:
             # Single-process data loading.
-            yield from [
+            yield from (
                 self.data_format_fn(bytes_content)
                 for bytes_content in dataflux_core.download.dataflux_download_lazy(
                     project_name=self.project_name,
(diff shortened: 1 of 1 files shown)


MattIrv (approved):
Would a possible approach for testing be to update the storage mock/fake to count the number of times that download is called, and verify that it's called the expected number of times prior to the whole set of data being returned, or something like that?

bernardhan33:
> Would a possible approach for testing be to update the storage mock/fake to count the number of times that download is called, and verify that it's called the expected number of times prior to the whole set of data being returned, or something like that?

Yeah the easier way is to work with fakes so that we don't have to mock out the entire return of "dataflux_download_lazy". Unfortunately we had to do that to avoid direct dependency call with storage_client.

PR #34, opened 2024-04-17
Presubmit Integration Testing

jdnurme:
Update presubmit testing to include full map and iterable integration tests. Please check and approve any open internal CLs prior to reviewing this, as kokoro cannot run until those changes are committed.

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/integration_tests/integration_test.py
+++ b/dataflux_pytorch/integration_tests/integration_test.py
@@ -69,6 +69,26 @@ class IntegrationTest(unittest.TestCase):
 
         return config
 
+    def train(self, data_loader, epochs, test_config):
+        training_start_time = time.time()
+        for i in range(epochs):
+            total_bytes = 0
+            for batch in data_loader:
+                # A simple sleep function to simulate the GPU training time.
+                time.sleep(0.1)
+
+                for object_bytes in batch:
+                    total_bytes += len(object_bytes)
+        training_end_time = time.time()
+        training_time = training_end_time - training_start_time
+        if (
+            test_config["download_timeout"]
+            and training_time > test_config["download_timeout"]
+        ):
+            raise AssertionError(
+                f"Expected download operation to complete in under {test_config['download_timeout']} seconds, but took {training_time} seconds."
+            )
+
     def test_list_and_load_iter(self):
         test_config = self.get_config()
         config = dataflux_iterable_dataset.Config()
@@ -103,24 +123,7 @@ class IntegrationTest(unittest.TestCase):
             persistent_workers=True,
             pin_memory=True,
         )
-        training_start_time = time.time()
-        for i in range(1):
-            total_bytes = 0
-            for batch in data_loader:
-                # A simple sleep function to simulate the GPU training time.
-                time.sleep(0.1)
-
-                for object_bytes in batch:
-                    total_bytes += len(object_bytes)
-        training_end_time = time.time()
-        training_time = training_end_time - training_start_time
-        if (
-            test_config["download_timeout"]
-            and training_time > test_config["download_timeout"]
-        ):
-            raise AssertionError(
-                f"Expected download operation to complete in under {test_config['download_timeout']} seconds, but took {training_time} seconds."
-            )
+        self.train(data_loader, 1, test_config)
 
     def test_list_and_load_map(self):
         test_config = self.get_config()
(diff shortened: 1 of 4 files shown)


dataflux_pytorch/tests/test_dataflux_checkpoint.py
```diff
@@ -14,6 +14,8 @@
  limitations under the License.
  """
 
+ # TEST CHANGE
+
```
jdnurme, line 18: Not certain how this was included, will remove.

dataflux_pytorch/integration_tests/integration_test.py
```diff
@@ -0,0 +1,193 @@
...
+        if config["download_timeout"]:
+            config["download_timeout"] = float(config["download_timeout"])
+        config["parallelization"] = (
+            int(config["parallelization"]) if config["parallelization"] else 1
+        )
+
+        return config
+
+    def test_list_and_load_iter(self):
+        list_start_time = time.time()
```
bernardhan33, line 67: A good chunk of the code here, such as defining the `data_format_fn` and the for-loop iteration that simulates the training are having some common code. Can we refactor those into some common helper code?

bernardhan33, line 67: nit: might be best to make the start time tight before the `dataset` creation

dataflux_pytorch/integration_tests/integration_test.py
```diff
@@ -0,0 +1,193 @@
...
+            prefetch_factor=2,
+            persistent_workers=True,
+            pin_memory=True,
+        )
+        training_start_time = time.time()
+        for i in range(1):
+            total_objects = 0
+            total_bytes = 0
+            epoch_start = time.time()
+            last_update = time.time()
```
bernardhan33, line 110: I don't think we need this now and we are not doing anything with it anyways

dataflux_pytorch/integration_tests/integration_test.py
```diff
@@ -0,0 +1,193 @@
...
+            dataset=dataset,
+            batch_size=100,
+            num_workers=test_config["parallelization"],
+            prefetch_factor=2,
+            persistent_workers=True,
+            pin_memory=True,
+        )
+        training_start_time = time.time()
+        for i in range(1):
+            total_objects = 0
```
bernardhan33, line 107: Do we want to add a check of matching total_objects?

dataflux_pytorch/integration_tests/integration_test.py
```diff
@@ -0,0 +1,193 @@
...
+            total_bytes = 0
```
bernardhan33, line 108: Similarly,

1. Do we want to add a check of matching total_bytes?
2. Have we confirmed that `len(object)` on a object type of Numpy array actually gives the size in bytes?

dataflux_pytorch/integration_tests/integration_test.py
```diff
@@ -0,0 +1,193 @@
...
+                total_objects += len(batch)
+                for object_bytes in batch:
+                    total_bytes += len(object_bytes)
+                if time.time() - last_update > 5:
+                    last_update = time.time()
+            epoch_end = time.time()
+        training_end_time = time.time()
+        if (
+            test_config["download_timeout"]
+            and training_time > test_config["download_timeout"]
```
bernardhan33, line 124: I don't think I see the `training_time` calculation in this function? I see it in the one below, though

dataflux_pytorch/integration_tests/integration_test.py
bernardhan33: I see it checked above. So maybe we don't need this. And this will be a little biased too since DataLoader can drop the last batch in the IterableDataset cases.

kokoro/presubmit.cfg
```diff
@@ -12,4 +12,66 @@
...
+  key: "TOTAL_FILE_SIZE"
+  value: "501770000"
+}
+
+env_vars {
+  key: "MAX_COMPOSE_BYTES"
+  value: "100000000"
+}
+
+env_vars {
```
bernardhan33, line 52: Nice to see that we are using the same timeouts as in dataflux-client-python! For listing, it should be mostly equivalent but for "download", using the same timeout will make the check tighter due to the various operations that DataLoader has and the `sleep_time`.

It's okay though. I just wanted to point out this implication.

kokoro/presubmit.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo Updating git submodules.
+    git submodule update
+}
+
+function install_requirements() {
+    echo Installing python3-pip.
+    sudo apt-get -y install python3-pip
+
+    echo Installing requirements.
+    pip install -r requirements.txt
```
bernardhan33, line 39: I don't think this is strictly needed as running `pip install .` will cover it, right? Have you seen problems without it?

dataflux_pytorch/integration_tests/integration_test.py
jdnurme: It does not return the correct value, so I'm leaving it out for now but intend to update it with a better check in the future. Currently just working out the kokoro specific kinks.

kokoro/presubmit.cfg
jdnurme: Yes if it exceeds the timeout or demonstrates inconsistency i'll extend it

kokoro/presubmit.sh
jdnurme: I did have problems without running both before, but it could have been one-off. I can test

PR #38, opened 2024-06-21
Catch exception when loading arrays from raw bytes fails

abhibyreddi:
The training workload crashes if an error occurs while loading numpy arrays from raw bytes. Catching those errors and moving on to the next file ensures that the workload finishes successfully.

### DIFF
--- a/demo/image-segmentation/data_loader.py
+++ b/demo/image-segmentation/data_loader.py
@@ -161,6 +161,12 @@ def get_data_loaders(flags, num_shards, global_rank):
 
 
 def collate_fn(batch):
+    """
+    Drop corrupt images that were previously
+    set to None during data loading step.
+    collate_fn is passed as an argument
+    to the DataLoader constructor above.
+    """
     batch = list(
         filter(lambda x: x["image"] is not None and x["label"] is not None, batch))
     return default_collate(batch)
--- a/demo/image-segmentation/data_loader.py
+++ b/demo/image-segmentation/data_loader.py
@@ -148,7 +148,7 @@ def get_data_loaders(flags, num_shards, global_rank):
 
     train_dataloader = DataLoader(
         train_dataset,
-        collate_fn=collate_fn
+        collate_fn=collate_fn,
         batch_size=flags.batch_size,
         shuffle=not flags.benchmark and train_sampler is None,
         sampler=train_sampler,
--- a/demo/image-segmentation/data_loader.py
+++ b/demo/image-segmentation/data_loader.py
@@ -20,6 +20,7 @@ import glob
 import numpy as np
 import torch
 from torch.utils.data import Dataset, DataLoader
+from torch.utils.data.dataloader import default_collate
 from torch.utils.data.distributed import DistributedSampler
 
 from dataflux_pytorch import dataflux_mapstyle_dataset
@@ -89,7 +90,8 @@ class SyntheticDataset(Dataset):
                 device=device,
                 requires_grad=False,
             )
-            self.y = torch.unsqueeze(self.y, dim=1 if layout == "NCDHW" else -1)
+            self.y = torch.unsqueeze(
+                self.y, dim=1 if layout == "NCDHW" else -1)
         else:
             y_shape = (
                 (channels_out,) + shape
@@ -146,6 +148,7 @@ def get_data_loaders(flags, num_shards, global_rank):
 
     train_dataloader = DataLoader(
         train_dataset,
+        collate_fn=collate_fn
         batch_size=flags.batch_size,
         shuffle=not flags.benchmark and train_sampler is None,
         sampler=train_sampler,
@@ -155,3 +158,9 @@ def get_data_loaders(flags, num_shards, global_rank):
     )
 
     return train_dataloader
+
+
+def collate_fn(batch):
+    batch = list(
+        filter(lambda x: x["image"] is not None and x["label"] is not None, batch))
+    return default_collate(batch)
--- a/demo/image-segmentation/data_loader.py
+++ b/demo/image-segmentation/data_loader.py
@@ -14,17 +14,20 @@
  limitations under the License.
  """
 
-import os
 import glob
+import os
 
 import numpy as np
 import torch
+
 from torch.utils.data import Dataset, DataLoader
 from torch.utils.data.dataloader import default_collate
+
+from pytorch_loader import DatafluxPytTrain
+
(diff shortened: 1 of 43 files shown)


demo/image-segmentation/pytorch_loader.py
```diff
@@ -282,9 +282,19 @@ def __getitems__(self, indices):
 
         res = []
         for i in range(len(images_in_bytes)):
+            try:
```
bernardhan33, line 285: It's ok to catch and skip for the purpose of completing the training successfully but it also will make the # data samples to be smaller than what's supposed to be. Is that desired?

bernardhan33: Also, I thought we ran into the same problem with the native way of loading datasets so that's going to be line 182. Do we need that to be modified?

bernardhan33: Finally, line 241 and 251 also does the data loading, do we need them modified too?

bernardhan33 (changes requested):

demo/image-segmentation/pytorch_loader.py
abhibyreddi: 1. I logged the names of the files that were causing the error. There weren't more than 3 to 4 problematic files. I believe that doesn't affect/minimally affects the metrics we are trying to collect (per-step training time) or the goal of this demo which is to show the benefits of dataflux (correct me if I'm wrong).

2, 3. The errors I ran into were with loading numpy arrays from raw bytes like here. Nevertheless I think it's good to wrap all calls to np.load in try-except.

bernardhan33: > 2, 3. The errors I ran into were with loading numpy arrays from raw bytes like here.

Yes this is expected because we defined the "__get_items__" method so PyTorch DataLoader will invoke this method. However, it will be good for the "__getitem__" method to incorporate this change. Similar for the native data loading.

abhibyreddi: Working on those changes! Although I'm not sure what to return when an exception occurs. Probably a random numpy array whose dimensions conform to the expectations of the training code. The dataloader currently has no knowledge of that so it's a bit tricky to make that change. Any recommendations?

abhibyreddi: Also, for the purposes of this demo here, I think it suffices to fix only `__getitems__` for now because the training code doesn't reference the files in the dataset by index. Neither does it ask the user to read files from a local directory. I could make changes to those two places for the sake of completeness but I believe it's not necessary as far as this demo is concerned. Happy to spend a little more time figuring out what to return from `__getitem__` when there's an error if you feel strongly that it makes sense to do it now.

bernardhan33: Probably should just raise the error instead.

`training code doesn't reference the files in the dataset by index` --> It still does. It just references a range of indices instead, when the `__getitems__` method is defined.

I would think that we should just keep all places consistent, for the sake of code quality. It would be confusing to see different data loading mechanisms, especially when the demo is instructing on how to **switch** from one dataset to the one with Dataflux.

MattIrv: There are some pretty extensive conversations on how to handle exceptions like this if you search for it, e.g. https://www.google.com/search?q=pytorch+data+loader+exceptions&oq=pytorch+d&gs_lcrp=EgZjaHJvbWUqCAgAEEUYJxg7MggIABBFGCcYOzIGCAEQRRg5MgYIAhBFGDwyBggDEEUYPDIGCAQQRRg8MgYIBRBFGEAyBggGEEUYQDIGCAcQRRhAqAIAsAIA&sourceid=chrome&ie=UTF-8. Is the approach here consistent with what those conversations are generally recommending?

MattIrv:
@abhibyreddi are you still planning to check this in?

bernardhan33:
Gentle ping on this

demo/image-segmentation/pytorch_loader.py
abhibyreddi: Followed https://github.com/pytorch/pytorch/issues/1137#issuecomment-618286571. Made changes to all `__getitem__` and `__getitems__` methods.

demo/image-segmentation/data_loader.py
```diff
@@ -153,3 +158,9 @@ def get_data_loaders(flags, num_shards, global_rank):
     )
 
     return train_dataloader
+
+
+def collate_fn(batch):
```
MattIrv, line 163: Can you add a comment explaining this?

demo/image-segmentation/pytorch_loader.py
```diff
@@ -186,12 +185,18 @@ def __len__(self):
...
-            "image": np.load(self.images[idx]),
-            "label": np.load(self.labels[idx])
-        }
-        data = self.rand_crop(data)
-        data = self.train_transforms(data)
+        try:
+            image = np.load(self.images[idx])
+        except:
+            image = None
+        try:
```
MattIrv, line 192: Does it make sense to have labels when the image is missing or vice-versa? Otherwise code like this might be better

```
try:
  image = np.load(self.images[idx])
  label = np.load(self.labels[idx])
except:
  image = None
  label = None
```

demo/image-segmentation/pytorch_loader.py
```diff
@@ -233,42 +238,55 @@ def __init__(
...
+            label = np.load(
+                io.BytesIO(
+                    dataflux_core.download.download_single(
+                        storage_client=self.storage_client,
+                        bucket_name=self.bucket_name,
+                        object_name=self.labels[idx][0],
+                    )
+                ),
+            )
+        except Exception as err:
```
MattIrv, line 282: Similar question as above, does it make sense to combine these try/except blocks?

demo/image-segmentation/pytorch_loader.py
```diff
@@ -311,4 +342,12 @@ def __len__(self):
...
 
     def __getitem__(self, idx):
-        return np.load(self.images[idx]), np.load(self.labels[idx])
+        try:
+            image = np.load(self.images[idx])
+        except:
+            image = None
+        try:
+            label = np.load(self.labels[idx])
+        except:
```
MattIrv, line 351: Similar comment as above

MattIrv (approved):

bernardhan33 (approved):

PR #40, opened 2024-06-24
Lightning checkpoint

divrawal:
Lightning Checkpoint support in dataflux for PyTorch. 

Includes 
- CheckpointIO interface implementation
- Unit tests
- Example file 
- Notebook
- README change.

- [x] Tests pass


dataflux_pytorch/examples/lightning_checkpoint.py
bernardhan33, line 1: We have a /demo folder for storing the demo code `simple_map_style_dataset.py` and `simple_iterable_dataset.py`. Shall we merge two directories?

pyproject.toml
```diff
@@ -29,6 +29,10 @@ dependencies = ["torch >= 2.0.1", "numpy", "google-cloud-storage", "absl-py"]
 [project.optional-dependencies]
 test = ["pytest"]
 
+lightning = [
```
bernardhan33, line 32: We put the dependencies in "requirements.txt" as well as in the "dependencies" field above. Can we move it there?

dataflux_pytorch/examples/lightning_checkpoint.py
```diff
@@ -0,0 +1,43 @@
...
+    dataset = WikiText2()
+    dataloader = DataLoader(dataset, num_workers=1)
+
+    model = LightningTransformer(vocab_size=dataset.vocab_size)
+    dataflux_ckpt = DatafluxLightningCheckpoint(project_name=project, bucket_name=bucket)
+    # Save once per step, and if `save_only_latest`, replace the last checkpoint each time.
+    # Replacing is implemented by saving the new checkpoint, and then deleting the previous one.
+    # If `save_only_latest` is False, a new checkpoint is created for each step.
+    checkpoint_callback = ModelCheckpoint(
+        save_top_k=-1 if save_only_latest else -1,
```
bernardhan33, line 18: This seems to be "-1" in both conditions.

dataflux_pytorch/examples/lightning_checkpoint.py
```diff
@@ -0,0 +1,43 @@
...
+        min_epochs=4,
+        max_epochs=5,
+        max_steps=3,
+        accelerator="cpu",
+    )
+    trainer.fit(model, dataloader)
+
+
+if __name__ == "__main__":
+    import os
```
bernardhan33, line 36: nit: group imports at top of file

dataflux_pytorch/tests/test_dataflux_lightning_checkpoint.py
```diff
@@ -0,0 +1,80 @@
+import torch
+import unittest
+
+from typing import Dict, Any
+from dataflux_client_python.dataflux_core.tests import fake_gcs
+from dataflux_pytorch.lightning.dataflux_lightning_checkpoint import DatafluxLightningCheckpoint
+
+LIGHTNING_ACCELERATOR = "cpu"
```
bernardhan33, line 8: nit: unused var

demo/lightning/Getting Started with Dataflux PyTorch Lightning Checkpoint with Google Cloud Storage.ipynb
bernardhan33, line 1: This is a little hard to review. Could you share a link to the Colab so we can view directly there?

bernardhan33 (commented):
Congrats on first contribution to dataflux-pytorch!

bernardhan33:
BTW the Kokoro test is failing, we will need to fix the test failure there.

demo/lightning/Getting Started with Dataflux PyTorch Lightning Checkpoint with Google Cloud Storage.ipynb
MattIrv: If you click on the "..." menu in the top-right of the file and then "view file" you can see the preview

dataflux_pytorch/examples/lightning_checkpoint.py
divrawal: I created a demo/notebook in that directory. Should I move this example code to be in the demo directory as well?

bernardhan33: yeah would be great to merge so we have a single place for hosting all the examples and demos

demo/lightning/Getting Started with Dataflux PyTorch Lightning Checkpoint with Google Cloud Storage.ipynb
```diff
@@ -0,0 +1,202 @@
...
+    "Then we need to install the rest of the rest of the expected packages for this demo."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "! pip install torch\n",
```
bernardhan33, line 34: nit: one-liner `pip install torch lightning`

demo/lightning/Getting Started with Dataflux PyTorch Lightning Checkpoint with Google Cloud Storage.ipynb
```diff
@@ -0,0 +1,202 @@
...
+    "## Lightning Checkpointing"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Dataflux PyTorch offers an optional implementation of PyTorch Lightning's checkpoints through implmentation of the CheckpointIO interface.\n",
+    "\n",
+    "The methods that are supported are `save_checkpoint`, `load_checkpoint`, `remove_checkpoint` and `teardown`. \n",
```
bernardhan33, line 75: `teardown` is basically a no-op. Should we still mention it?

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
```diff
@@ -0,0 +1,77 @@
...
+            self.storage_client = storage.Client(
+                project=self.project_name,
+                client_info=ClientInfo(user_agent="dataflux/0.0"),
+            )
+        self.bucket = self.storage_client.bucket(self.bucket_name)
+
+    def _parse_gcs_path(self, path: str) -> str:
+        if not path or not path.startswith("gcs://"):
+            raise ValueError("Path needs to begin with gcs://")
+        path = path[len("gcs://") :]
```
bernardhan33, line 30: Oh I just realized that we are using `gcs://` as the prefix. I feel that this is okay for now but just a heads up that gcsfs and GCS APIs (such as gsutil and gcloud cp, etc) are paths starting with `gs://`. Not sure if we want to make the conventions uniform

demo/lightning/Getting Started with Dataflux PyTorch Lightning Checkpoint with Google Cloud Storage.ipynb
bernardhan33, line 1: nit: Seems that the cells here are not run. We can run the cells and save the outputs along with the cells to show the results (and optionally explain the results).

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
divrawal: I think [gcsfs](https://gcsfs.readthedocs.io/en/latest/#integration) uses `gcs://` as well. Lightning documentation also confirms this https://lightning.ai/docs/pytorch/stable/common/remote_fs.html 

If there is an edit I think it is to support `gs://` paths here, I can go ahead and add that as an allowable prefix?

bernardhan33: Oh right. It supports both `gcs://` and `gs://`. Yeah it would be great to allow this flexibility. Thanks!

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
```diff
@@ -0,0 +1,77 @@
+import torch
+from typing import Optional, Dict, Any
+
+from lightning.pytorch.plugins.io import CheckpointIO
+from google.cloud import storage
+from google.api_core.client_info import ClientInfo
+
+class DatafluxLightningCheckpoint(CheckpointIO):
+    """A checkpoint manager for GCS using the :ckass:'CheckpointIO' interface"""
```
MattIrv, line 9: Typo here in :ckass: ?

dataflux_client_python
MattIrv, line 1: @bernardhan33 is this ok? Do we have any specific process for updating the version of the dataflux_client_python repo in the git submodule?

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
```diff
@@ -0,0 +1,77 @@
...
+        self.bucket_name = bucket_name
+        self.storage_client = storage_client
+        if not storage_client:
+            self.storage_client = storage.Client(
+                project=self.project_name,
+                client_info=ClientInfo(user_agent="dataflux/0.0"),
+            )
+        self.bucket = self.storage_client.bucket(self.bucket_name)
+
+    def _parse_gcs_path(self, path: str) -> str:
```
MattIrv, line 27: Since the client is initialized with a project name and bucket name, why is this necessary? Could we just have the user provide a path which is relative to the bucket rather than starting with "gcs://<bucket>/"?

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
```diff
@@ -0,0 +1,77 @@
...
+        key = self._parse_gcs_path(path)
+        blob = self.bucket.blob(key)
+        return torch.load(blob.open("rb"), map_location)
+
+    def remove_checkpoint(
+        self,
+        path: str,
+    ) -> None:
+       key = self._parse_gcs_path(path)
+       """This command may take a while if it is many objects
```
MattIrv, line 69: Is this comment relevant? It looks like it only supports a single object

MattIrv (approved):

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def save_checkpoint(
+        self,
+        checkpoint: Dict[str, Any],
+        path: str,
+        storage_options: Optional[Any] = None,
+    ) -> None:
+        key = self._parse_gcs_path(path)
+        blob = self.bucket.blob(key)
+        return torch.save(checkpoint, blob.open("wb", ignore_flush=True))
```
MattIrv, line 53: Do the blob.open calls here or on line 62 need to be closed/flushed at all? From the [docs](https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.blob.Blob#google_cloud_storage_blob_Blob_open) I see "This method can be used as a context manager, just like Python's built-in 'open()' function." and the provided [code sample](https://github.com/googleapis/python-storage/blob/main/samples/snippets/storage_fileio_write_read.py) uses it like `with blob.open("w") as f:`

dataflux_client_python
bernardhan33: Yes this is the right thing to do. Although definitely feasible, we don't have any automatic processes now. We have a tracking ticket (b/325531246) to do this where the submodules can get constantly updated.

However, manual updates are still needed in PRs like this to make sure any additional dependencies of the submodules get checked.

When we upload the package to PyPI it will grab the contents in the submodule here. I'll demo this next week once this gets merged.

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
divrawal: That was the original implementation but after I wrote the demo/example file I saw that for automated checkpointing this needed to ingest a full path. The full path is what gets passed through to the trainer through the ModelCheckpoint object.

MattIrv: Why can't we update the demo to not pass in the "gcs://<bucket>/" part of the path? Alternatively, could we remove the bucket from the constructor in this file? It seems duplicative as-is

bernardhan33: Yeah it sounds like we are taking on [this feature](https://docs.google.com/document/d/1GVC3bsTGQif03Ps2qFqKRJxxaPFmCcAQoSgL6oRx62o/edit?resourcekey=0-XFc1Vt8Vx4Xl4b8qskVyUQ&tab=t.pxu2qfhdvv07#heading=h.d7bfmv2o8k3y) essentially, which I think it's great to have to make the user experience similar with gcsfs, etc. I'd also agree on just taking away buckets from the constructor. We'd also need work to support this for the two datasets though but definitely for different PRs.

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
divrawal: Looking into this

divrawal: Changed the implementation to context manager, which is more in style and better practice. Had to update the fake in the client code so opening another PR for that now.

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
MattIrv: I don't see any code changes here, am I missing them? @divrawal can you follow up on it?

PR #43, opened 2024-06-28
Attempt to fix the continuous build failure by introducing virtual environment

bernardhan33:
From the [error message](https://fusion2.corp.google.com/invocations/e3a81ae2-9bb1-4404-948a-3a4c81020846/targets/cloud-storage-ml-infra-dataflux%2Fdataflux-pytorch%2Fcontinuous), there are

```
Attempting uninstall: PyYAML
    Found existing installation: PyYAML 5.3.1
```

while the lightning dependency is pulling 6.0.1

```
Collecting PyYAML<8.0,>=5.4 (from lightning>=2.0->gcs-torch-dataflux==1.0.0)
  Downloading PyYAML-6.0.1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (2.1 kB)
```

Therefore I'm suspecting that this dependency was present in the original running environment and we never used virtual environment to set up the test environment.

### DIFF
--- a/kokoro/build.sh
+++ b/kokoro/build.sh
@@ -20,6 +20,12 @@ set -e
 # Code under repo is checked out to this directory.
 cd "${KOKORO_ARTIFACTS_DIR}/github/dataflux-pytorch"
 
+function setup_virtual_envs() {
+    echo Setting up Python virtual environment.
+    python3 -m venv venv
+    source venv/bin/activate
+}
+
 function run_git_commands() {
     echo Setting git permissions.
     git config --global --add safe.directory "*" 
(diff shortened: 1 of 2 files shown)


kokoro/build.sh
```diff
@@ -20,6 +20,13 @@ set -e
 # Code under repo is checked out to this directory.
 cd "${KOKORO_ARTIFACTS_DIR}/github/dataflux-pytorch"
 
+function setup_virtual_envs() {
+    echo Setting up Python virtual environment.
+    sudo apt install python3.8-venv
```
bernardhan33, line 25: Tried "python3-venv" but can't find candidate. Did some searches and seemed that we needed to specify the python version

MattIrv (approved):
Thanks Bernard! Any idea why this wasn't caught by the presubmit test run?

bernardhan33:
> Thanks Bernard! Any idea why this wasn't caught by the presubmit test run?

Ah tried to think through this but don't have a clue. Thought it might be due to different machine hosts but we always ran this with docker.

PR #44, opened 2024-07-03
Add threaded download to map-style dataset.

MattIrv:
I'm not seeing it improve or decrease performance unfortunately, but it's worth adding so that we can test it on some different use cases.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR (none)

### DIFF
--- a/dataflux_pytorch/dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/dataflux_mapstyle_dataset.py
@@ -52,12 +52,14 @@ class Config:
         num_processes: int = os.cpu_count(),
         prefix: str = None,
         max_listing_retries: int = 3,
+        threads_per_process: int = 1,
     ):
         self.sort_listing_results = sort_listing_results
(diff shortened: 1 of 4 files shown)


MattIrv:
Seeing that threading doesn't make a difference on larger files (3 MB) in my benchmarks, but makes a large difference on 10KB files:

500000 10KB files:
With 16 threads per worker
Epoch 0 took 31.135604858398438 seconds to iterate over 500000 objects and 5000000000 bytes.
All training (1 epochs) took 31.135690212249756 seconds.

With 1 thread per worker
Epoch 0 took 410.33694410324097 seconds to iterate over 500000 objects and 5000000000 bytes.
All training (1 epochs) took 410.3370113372803 seconds.

There are some warnings that I want to investigate in a follow-up:
"WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10"

bernardhan33 (approved):

PR #46, opened 2024-07-10
add disable_compose config

jdnurme:
Add explicit flag to disable compose. High scale operations must disable compose to avoid errors due to high read/write load caused by compose calls.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,8 +325,7 @@ To optimize the download performance of small files, the Accelerated Dataloader 
 gcloud storage rm --recursive gs://<my-bucket>/dataflux-composed-objects/
 ```
 
-You can also turn off this behavior by setting the “max_composite_object_size” parameter to 0 when constructing the dataset. Example:
-
+You can turn of this behavior by setting the "disable_compose" parameter to true, or by setting the “max_composite_object_size” parameter to 0 when constructing the dataset. Example:
 ```python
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -325,20 +325,22 @@ To optimize the download performance of small files, the Accelerated Dataloader
...
   config=dataflux_mapstyle_dataset.Config(
     prefix=PREFIX,
     max_composite_object_size=0,
+    disable_compose=True,
   ),
 )
 ```
 
-Note that turning off this behavior will cause the training loop to take significantly longer to complete when working with small files.
+**Note that this feature should always be disabled for highly scaled operations that are likely to approach the maximum throughput speeds allowed in the user project.** If left enabled on such operations, users are likely to hit their throughput limit earlier than expected.
```
MattIrv, line 341: Maybe a bit lighter wording would be a good idea here, at least until the guide on 429 errors is in place, something like "Note that turning off this behavior may cause the training loop to take significantly longer to complete when working with small files. However, composed download will hit QPS limits at a lower scale than downloading files directly, so you should disable this behavior when running at high multi-node scales where you are able to hit project QPS or throughput limits without using composed download."

MattIrv (approved):

bernardhan33 (approved):

PR #48, opened 2024-07-11
update readme with 429 info

jdnurme:
Briefly cover 429 errors with troubleshooting steps and actions.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,13 +353,15 @@ Many machine learning efforts opt for a highly distributed training model levera
 #### Egress Throughput Limits
 429 errors acompanied with messages indicating `This workload is drawing too much egress bandwidth from Google Cloud Storage` or `triggered the Cloud Storage Egress Bandwidth Cap` indicate that the data throughput rate of your workload is exceeding the maximum capacity of your Google Cloud Project. To address these issues, you can take the following steps:
 
-1. Ensure that object composition is [disabled](#composite-objects) for dataflux download operations.  
-2. Check that other workloads executing within your project are not drawing excess bandwidth.
-3. Lower the scale of the workload in question to reduce its egress bandwidth (e.g. fewer running pods).
-4. Contact your Technical Account Manager or Google representative to file a Cloud Capacity Advisor request on your behalf. This request should specify that you wish to increase the bandwidth caps on your project.
+1. Check that other workloads executing within your project are not drawing excess bandwidth. Bandwidth usage can be viewed by following steps [here](https://cloud.google.com/storage/docs/bandwidth-usage#bandwidth-monitoring).
+2. Follow [these instructions](https://cloud.google.com/storage/docs/bandwidth-usage#increase) to apply for a quota increae of up to 1tbps.
+3. If more than 1Tpbs is required, contact your Technical Account Manager or Google representative to file a request on your behalf. This request should specify that you wish to increase the bandwidth caps on your project.
+4. Try [disabling](#composite-objects) object composition for Dataflux download operations. While object composition increases performance on small files, it also increases the egress bandwidth load when compared to standard download. If you are consistently hitting the bandwidth limit, disabling this feature can improve the performance of retry-backoff operations.
 
 #### QPS Limits
-QPS limits can trigger 429 errors, but more commonly manifest in slower than expected execution times. QPS bottlenecks are more common when operating on high volumes of small files. Note that bucket QPS limits will naturally scale overtime, so allowing a grace period for warmup can often lead to faster performance. To get more detail on the performance of a target bucket, look at the `Observability` tab when viewing your bucket from the Cloud Console.
+QPS limits can trigger 429 errors with a body message indicating `Too many Requests`, but more commonly manifest in slower than expected execution times. QPS bottlenecks are more common when operating on high volumes of small files. Note that bucket QPS limits will [naturally scale over time](https://cloud.google.com/storage/docs/request-rate#best-practices), so allowing a grace period for warmup can often lead to faster performance. To get more detail on the performance of a target bucket, look at the `Observability` tab when viewing your bucket from the Cloud Console.
+
+If your workload is failing with messages similar to `TooManyRequests: 429 GET https://storage.googleapis.com/download/storage/v1/b/<MY-PROJECT>/o/dataflux-composed-objects%2Fa80727ae-7bc9-4ba3-8f9b-13ff001d6418` that contain the "dataflux-composed-objects" keyword, [disabling](#composite-objects) composed objects is the best first troubleshooting step. This can reduce QPS load brought on by the compose API when used at scale.
 
 ## Contributing
 We welcome your feedback, issues, and bug fixes. We have a tight roadmap at this time so if you have a major feature or change in functionality you'd like to contribute, please open a GitHub Issue for discussion prior to sending a pull request. Please see [CONTRIBUTING](docs/contributing.md) for more information on how to report bugs or submit pull requests.
--- a/README.md
+++ b/README.md
@@ -353,9 +353,10 @@ Many machine learning efforts opt for a highly distributed training model levera
 #### Egress Throughput Limits
 429 errors acompanied with messages indicating `This workload is drawing too much egress bandwidth from Google Cloud Storage` or `triggered the Cloud Storage Egress Bandwidth Cap` indicate that the data throughput rate of your workload is exceeding the maximum capacity of your Google Cloud Project. To address these issues, you can take the following steps:
 
-1. Check that other workloads executing within your project are not drawing excess bandwidth.
-2. Lower the scale of the workload in question to reduce its egress bandwidth (e.g. fewer running pods).
-3. Contact your Technical Account Manager or Google representative to file a Cloud Capacity Advisor request on your behalf. This request should specify that you wish to increase the bandwidth caps on your project.
+1. Ensure that object composition is [disabled](#composite-objects) for dataflux download operations.  
+2. Check that other workloads executing within your project are not drawing excess bandwidth.
+3. Lower the scale of the workload in question to reduce its egress bandwidth (e.g. fewer running pods).
+4. Contact your Technical Account Manager or Google representative to file a Cloud Capacity Advisor request on your behalf. This request should specify that you wish to increase the bandwidth caps on your project.
 
 #### QPS Limits
 QPS limits can trigger 429 errors, but more commonly manifest in slower than expected execution times. QPS bottlenecks are more common when operating on high volumes of small files. Note that bucket QPS limits will naturally scale overtime, so allowing a grace period for warmup can often lead to faster performance. To get more detail on the performance of a target bucket, look at the `Observability` tab when viewing your bucket from the Cloud Console.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -347,6 +347,20 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
+
+1. Ensure that object composition is [disabled](#composite-objects) for dataflux download operations.  
+2. Check that other workloads executing within your project are not drawing excess bandwidth.
+3. Lower the scale of the workload in question to reduce its egress bandwidth (e.g. fewer running pods).
+4. Contact your Technical Account Manager or Google representative to file a Cloud Capacity Advisor request on your behalf. This request should specify that you wish to increase the bandwidth caps on your project.
+
+#### QPS Limits
+QPS limits can trigger 429 errors, but more commonly manifest in slower than expected execution times. QPS bottlenecks are more common when operating on high volumes of small files. Note that bucket QPS limits will naturally scale overtime, so allowing a grace period for warmup can often lead to faster performance. To get more detail on the performance of a target bucket, look at the `Observability` tab when viewing your bucket from the Cloud Console.
```
bernardhan33, line 362: Could we link "naturally scale overtime" with https://cloud.google.com/storage/docs/request-rate#best-practices?

README.md
```diff
@@ -347,6 +347,20 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
 
+### Throughput and QPS Limits
+Many machine learning efforts opt for a highly distributed training model leveraging tools such as Pytorch Lightning and Ray. These models are compatible with Dataflux, but can often trigger the rate limits of Google Cloud Storage. This typically manifest in a 429 error, or slower than expected speeds while running distributed operations. Details on specific quotas and limits can be found [here](https://cloud.google.com/storage/quotas).
+
+#### Egress Throughput Limits
+429 errors acompanied with messages indicating `This workload is drawing too much egress bandwidth from Google Cloud Storage` or `triggered the Cloud Storage Egress Bandwidth Cap` indicate that the data throughput rate of your workload is exceeding the maximum capacity of your Google Cloud Project. To address these issues, you can take the following steps:
+
+1. Ensure that object composition is [disabled](#composite-objects) for dataflux download operations.  
+2. Check that other workloads executing within your project are not drawing excess bandwidth.
+3. Lower the scale of the workload in question to reduce its egress bandwidth (e.g. fewer running pods).
```
bernardhan33, line 358: I'm a little hesitant on whether it makes sense to suggest customers to "lower the scale" to work around those problems. I mean, it could make sense to do that to validate the problem but as a solution, I'm not sure?

bernardhan33 (approved):

README.md
```diff
@@ -347,6 +347,20 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
 
+### Throughput and QPS Limits
+Many machine learning efforts opt for a highly distributed training model leveraging tools such as Pytorch Lightning and Ray. These models are compatible with Dataflux, but can often trigger the rate limits of Google Cloud Storage. This typically manifest in a 429 error, or slower than expected speeds while running distributed operations. Details on specific quotas and limits can be found [here](https://cloud.google.com/storage/quotas).
+
+#### Egress Throughput Limits
+429 errors acompanied with messages indicating `This workload is drawing too much egress bandwidth from Google Cloud Storage` or `triggered the Cloud Storage Egress Bandwidth Cap` indicate that the data throughput rate of your workload is exceeding the maximum capacity of your Google Cloud Project. To address these issues, you can take the following steps:
+
+1. Ensure that object composition is [disabled](#composite-objects) for dataflux download operations.
```
MattIrv, line 356: Could you add a bullet point explaining this recommendation? In fact I'm not sure it makes sense here - if you're hitting egress throughput limits, then you might still hit them if you disable compose, but maybe not.

README.md
```diff
@@ -347,6 +347,20 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
 
+### Throughput and QPS Limits
+Many machine learning efforts opt for a highly distributed training model leveraging tools such as Pytorch Lightning and Ray. These models are compatible with Dataflux, but can often trigger the rate limits of Google Cloud Storage. This typically manifest in a 429 error, or slower than expected speeds while running distributed operations. Details on specific quotas and limits can be found [here](https://cloud.google.com/storage/quotas).
+
+#### Egress Throughput Limits
+429 errors acompanied with messages indicating `This workload is drawing too much egress bandwidth from Google Cloud Storage` or `triggered the Cloud Storage Egress Bandwidth Cap` indicate that the data throughput rate of your workload is exceeding the maximum capacity of your Google Cloud Project. To address these issues, you can take the following steps:
+
+1. Ensure that object composition is [disabled](#composite-objects) for dataflux download operations.  
+2. Check that other workloads executing within your project are not drawing excess bandwidth.
```
MattIrv, line 357: Could we give any guidance on how to do this, such as referring to Cloud Monitoring graphs or any existing documentation on how to monitor GCS bandwidth?

README.md
```diff
@@ -347,6 +347,20 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
+3. Lower the scale of the workload in question to reduce its egress bandwidth (e.g. fewer running pods).
+4. Contact your Technical Account Manager or Google representative to file a Cloud Capacity Advisor request on your behalf. This request should specify that you wish to increase the bandwidth caps on your project.
```
MattIrv, line 359: There are no results when I search ['gcp "Cloud Capacity Advisor"'](https://www.google.com/search?q=gcp+%22cloud+capacity+advisor%22&sca_esv=1fac4576b0bfa454&sxsrf=ADLYWIJ2mUa4QImCbFmvezANRtkuyzJD5Q%3A1720729464125&ei=eD-QZqCmB42B0PEP5-q_6AM&ved=0ahUKEwigwtrj6J-HAxWNADQIHWf1Dz0Q4dUDCA8&uact=5&oq=gcp+%22cloud+capacity+advisor%22&gs_lp=Egxnd3Mtd2l6LXNlcnAiHGdjcCAiY2xvdWQgY2FwYWNpdHkgYWR2aXNvciIyCBAhGKABGMMESKAFUNIBWJAEcAJ4AJABAJgBV6AB8QGqAQEzuAEDyAEA-AEBmAICoAKjAZgDAIgGAZIHATKgB8wI&sclient=gws-wiz-serp) on Google. Is there some better way we can refer to this or any existing documentation we can link?

README.md
MattIrv, line 362: nit: "over time" not "overtime"

README.md
```diff
@@ -347,6 +347,20 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
+
+#### QPS Limits
```
MattIrv, line 361: Can you provide sample error messages and recommendations for this case like you did above? Specifically for when to disable Composed Download?

README.md
MattIrv, line 356: nit: capitalize Dataflux to be consistent with the rest of the doc

jdnurme: You could definitely still hit them, however they are much more retry-able if its on a regular download. Due to the extra load caused by compose, the backoff tends to progress much further when trying to compose before it can resume.

README.md
jdnurme: I pulled this name directly from the error message, so it could be a legacy term. I'll remove it.

README.md
MattIrv: Maybe this is in the wrong section then, won't that give a QPS error rather than an egress throughput error?

README.md
MattIrv: Oh, it could be ok then, but to avoid confusion you might just say something like "capacity request". In the second sentence, it would probably be good to specify "GCS bandwidth cap" also.

README.md
```diff
@@ -347,6 +347,22 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
+
+#### Egress Throughput Limits
+429 errors acompanied with messages indicating `This workload is drawing too much egress bandwidth from Google Cloud Storage` or `triggered the Cloud Storage Egress Bandwidth Cap` indicate that the data throughput rate of your workload is exceeding the maximum capacity of your Google Cloud Project. To address these issues, you can take the following steps:
+
+1. Check that other workloads executing within your project are not drawing excess bandwidth. Bandwidth usage can be viewed by following steps [here](https://cloud.google.com/storage/docs/bandwidth-usage#bandwidth-monitoring).
+2. Follow [these instructions](https://cloud.google.com/storage/docs/bandwidth-usage#increase) to apply for a quota increae of up to 1tbps.
+3. If more than 1Tpbs is required, contact your Technical Account Manager or Google representative to file a request on your behalf. This request should specify that you wish to increase the bandwidth caps on your project.
```
MattIrv, line 358: Typo, "1Tpbs:. Besides the typo, it will be good to have the capitalization be consistent with line 2 also (and with the linked docs)

README.md
```diff
@@ -347,6 +347,22 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
+4. Try [disabling](#composite-objects) object composition for Dataflux download operations. While object composition increases performance on small files, it also increases the egress bandwidth load when compared to standard download. If you are consistently hitting the bandwidth limit, disabling this feature can improve the performance of retry-backoff operations.
```
MattIrv, line 359: This recommendation doesn't make sense to me. If we're hitting the throughput cap with compose enabled, then disabling compose can only lower the throughput. You may not hit the error, but that's because your throughput will be lower.

README.md
jdnurme: I put it as a low priority troubleshooting step in this segment since it could help, but I put it as the primary troubleshooting step for the QPS error use case listed below if that seems ok. I can fully remove it if thats preferred tho.

README.md
```diff
@@ -347,6 +347,22 @@ To avoid storage charges for retaining the temporary composite objects, consider
...
+
+#### QPS Limits
+QPS limits can trigger 429 errors with a body message indicating `Too many Requests`, but more commonly manifest in slower than expected execution times. QPS bottlenecks are more common when operating on high volumes of small files. Note that bucket QPS limits will [naturally scale over time](https://cloud.google.com/storage/docs/request-rate#best-practices), so allowing a grace period for warmup can often lead to faster performance. To get more detail on the performance of a target bucket, look at the `Observability` tab when viewing your bucket from the Cloud Console.
+
+If your workload is failing with messages similar to `TooManyRequests: 429 GET https://storage.googleapis.com/download/storage/v1/b/<MY-PROJECT>/o/dataflux-composed-objects%2Fa80727ae-7bc9-4ba3-8f9b-13ff001d6418` that contain the "dataflux-composed-objects" keyword, [disabling](#composite-objects) composed objects is the best first troubleshooting step. This can reduce QPS load brought on by the compose API when used at scale.
```
MattIrv, line 364: You may want to mention that the user should monitor throughput to make sure this doesn't cause an overall decrease in throughput. If it does, then the better solution is to make sure the retry is configured correctly to ignore and retry the errors without failing.

MattIrv (approved):

README.md
jdnurme: So I'm targeting the retry of composed object downloads. It seemed from testing that the retry would max out when grabbing multiple composed downloads, but wouldn't error out when retrying on non-composed files. The issues i was encountering were surrounding any time the compose operations retried. It seems like they would have to cycle down further than normal download. Both of these errors occurred while throughput was bouncing off of the ~4.5 Tbps mark. The issue seems to occur regardless of what causes the composed object to retry. I'll just remove the suggestion for now since its more of a bug with retry, and they can always follow the suggestion under the QPS section.

README.md
jdnurme: I'm going to include that instruction with the CL that allows the configuration of the params.

PR #53, opened 2024-07-12
Standardize Python extensions and formatting settings.

MattIrv:
This will make sure we're all using the same formatter and settings.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,20 @@
 {
-    "editor.formatOnSave": true,
+    "[python]": {
+        "editor.formatOnSaveMode": "file",
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "eeyore.yapf",
+        "editor.formatOnType": false,
+        "files.trimTrailingWhitespace": true,
+    },
     "python.testing.pytestArgs": [
         "-s",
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,9 +1,20 @@
 {
-    "editor.formatOnSave": true,
+    "[python]": {
+        "editor.formatOnSaveMode": "file",
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "eeyore.yapf",
```
bernardhan33, line 5: any particular reason that you chose the "eeyore.yapf" formatter? For the more famous [black formatter](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter) I have seen inconvenience such as setting tab space, etc so I'd like to try out yapf as it seems to support indentation space, etc.

bernardhan33 (approved):

.vscode/settings.json
MattIrv: According to https://github.com/google/yapf?tab=readme-ov-file#formatting-style yapf is preferred for google open source, though I’m not sure how official that is.

bernardhan33: Sounds good. I just needed the indentation customization so that formatter looks good to me.

PR #57, opened 2024-07-15
Add boilerplate code for Dataflux-Pytorch Lightning demo

abhibyreddi:


demo/lightning/data.py
```diff
@@ -0,0 +1,24 @@
+import torch
+import lightning.pytorch as pl
+
+from dataflux_pytorch import dataflux_mapstyle_dataset
+
+
+class Unet3DDataModule(pl.LighitningDataModule):
+    def __init__(self, gcs_bucket):
+        self.data_dir = gcs_bucket + "/images"
```
bernardhan33, line 9: Use os.path.join() for directory joins.

bernardhan33 (approved):

demo/lightning/data.py
MattIrv, line 9: Should this path be configurable, with this as the default?

demo/lightning/train.py
```diff
@@ -0,0 +1,14 @@
...
+from model import Unet3D
+from data import Unet3DDataModule
+
+if __name__ == "__main__":
+    model = Unet3D()
+    train_data_loader = Unet3DDataModule()
+    trainer = pl.Trainer(
+        devices=2, 
+        accelerator="gpu", 
+        max_epochs=5
```
MattIrv, line 12: I think these should all be configurable too

demo/lightning/data.py
MattIrv, line 1: Regarding the file path, JD is working on a similar demo for text-based training now, so maybe this should live in its own sub-folder within demo/lightning?

MattIrv (approved):

abhibyreddi (commented):
Moved the three new files from demo/lightning to demo/lightning/image-segmentation. Will probably have to move demo/image-segmentation to demo/pytorch/image-segmentation in a later PR.

I agree that the arguments to Unet3D and Unet3DDataModule constructors should be configurable. I will make that change after I have this workload up and running with hardcoded parameters.

PR #59, opened 2024-07-16
Refactor Dataflux simple demo loops and add retry flags

MattIrv:
A few things in this PR:

* Add retry flags to the demo loop code now that it's configurable
* Add a base-level Dockerfile to containerize the whole repo for testing on GKE
* Restructure the demos and give them GKE deployment files, along with a README
* Other minor fixes in the demo loops

- [x] Tests pass

### DIFF
--- a/demo/list-and-download/iterable/deployment.yaml
+++ b/demo/list-and-download/iterable/deployment.yaml
@@ -1,0 +1,55 @@
+# GKE deployment script for the Dataflux container.
+# If you push a new version of the container, you need to update the image.
+# You can set the environment variables from main.go to control the flags.
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: mirvine-dataflux
+  namespace: default
+spec:
+  backoffLimit: 6
+  completionMode: NonIndexed
+  completions: 100
+  manualSelector: false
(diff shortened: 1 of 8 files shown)


demo/list-and-download/iterable/deployment.yaml
```diff
@@ -0,0 +1,53 @@
+# GKE deployment script for the Dataflux iterable-style demo loop.
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: mirvine-dataflux
```
abhibyreddi, line 5: Is it ok for this to be same for all runs of the demo or will it cause any conflicts in rare cases where someone attempts to run this demo when a previous run is in progress?

abhibyreddi (approved):

demo/list-and-download/iterable/deployment.yaml
MattIrv: You need to delete any existing runs with the same name, even completed, before you can kick off a new one. I want to look into it a bit more next, but I don't see any really convenient way to pass things like the job name or flags as parameters with Kubernetes unfortunately.

divrawal (approved):

PR #60, opened 2024-07-16
Implement data module for the pytorch lightning workload

abhibyreddi:
Reuse the DatafluxPytTrain class from the original demo & move the transform functions to the data module definition.


demo/lightning/image-segmentation/data.py
```diff
@@ -1,23 +1,144 @@
...
+from pytorch_loader import DatafluxPytTrain
 
 
 class Unet3DDataModule(pl.LighitningDataModule):
-    def __init__(self, gcs_bucket, prefix):
-        self.data_dir = os.path.join(gcs_bucket, prefix)
+
+    def __init__(self, args):
+        super().__init__()
+        self.args = args
```
MattIrv, line 33: Is there some specific reason this is taking an untyped args object rather than individual arguments one-by-one? If so, can you add a comment explaining it?

demo/lightning/image-segmentation/data.py
```diff
@@ -1,23 +1,144 @@
...
+                self.train_sampler = DistributedSampler(
+                    self.train_dataset, 
+                    seed=self.args.seed, 
+                    drop_last=True)
 
     def train_dataloader(self):
-        # Init dataflux dataloader
-        # Wrap it in torch.DataLoader
-        pass
+        return  DataLoader(
```
MattIrv, line 62: nit: There is a double space here. Have you run the formatter? If not, you should be able to install it from the Extensions > Recommended Extensions panel on VS Code and then it will run on save

demo/lightning/image-segmentation/data.py
```diff
@@ -1,23 +1,144 @@
...
+            batch_size=self.args.batch_size,
+            shuffle=not self.args.benchmark and self.train_sampler is None,
+            sampler=self.train_sampler,
+            num_workers=self.args.num_dataloader_threads,
+            pin_memory=True,
+            drop_last=True,
+        )
+
+
+def get_train_transforms():
```
MattIrv, line 73: Can you add a comment explaining this function? In general most of the functions/methods/types in this PR would benefit from doc comments

MattIrv (approved):

demo/lightning/image-segmentation/data.py
```diff
@@ -1,23 +1,144 @@
...
-        # Wrap it in torch.DataLoader
-        pass
+        return  DataLoader(
+            self.train_dataset,
+            batch_size=self.args.batch_size,
+            shuffle=not self.args.benchmark and self.train_sampler is None,
+            sampler=self.train_sampler,
+            num_workers=self.args.num_dataloader_threads,
+            pin_memory=True,
+            drop_last=True,
```
bernardhan33, line 69: persistent_worker should be set too.

demo/lightning/image-segmentation/data.py
bernardhan33: +1.

demo/lightning/image-segmentation/pytorch_loader.py
```diff
@@ -0,0 +1,203 @@
...
+        self.bucket_name = bucket_name
+        self.config = config
+        self.dataflux_download_optimization_params = (
+            dataflux_core.download.DataFluxDownloadOptimizationParams(
+                max_composite_object_size=self.config.max_composite_object_size
+            ))
+        if not storage_client:
+            self.storage_client = storage.Client(
+                project=project_name,
+                client_info=ClientInfo(user_agent="dataflux/0.0"),
```
bernardhan33, line 59: client_info needs not to be manually set now, but just a client passed in. https://github.com/GoogleCloudPlatform/dataflux-client-python/blob/main/dataflux_core/fast_list.py#L176

bernardhan33 (approved):

demo/lightning/image-segmentation/data.py
abhibyreddi: `args` has multiple key-value pairs which are needed by the other methods of this class. It will be constructed from the command line arguments. I will have to re-use the code in `demo/image-segmentation/arguments.py` to do that. This will be implemented in the next PR.

demo/lightning/image-segmentation/data.py
abhibyreddi: Most of the code in this PR is copied from the original demo and has been refactored to conform to pytorch lightning's abstractions. My plan was to build out the demo piece by piece (implement the data module class and then the actual module etc.) then refine it further. Is it okay if I implement all the classes fully and then add comments in the final PR?

demo/lightning/image-segmentation/pytorch_loader.py
abhibyreddi: Removed here and in `demo/image-segmentation/pytorch_loader.py`

PR #61, opened 2024-07-17
Updated speeds to reflect new numbers when benchmarked with the same …

divrawal:
Updated table with benchmark numbers to reflect when timings were done in the same region (source/bucket).

- [x] Tests pass


dataflux_pytorch/benchmark/README.md
```diff
@@ -94,9 +94,9 @@ Dataflux's implementation of CheckpointIO for PyTorch Lightning is undergoing ac
    </td>
```
MattIrv, line 94: Could you also add a description at the top of the table about the VM size you were using for this, and the fact that it and the bucket are in the same region? So that anyone who sees these benchmarks and tries to reproduce them doesn't hit any issues with those things

MattIrv (approved):

Issue #58, opened 2024-07-16
Pickling client objects is explicitly not supported (Iterable Dataset)


SkanderBS2024:
Hello,

I've been running the following function which loads data from a bucket containg (1 txt and 2csv files)

<img width="880" alt="Capture d’écran 2024-07-16 à 14 45 10" src="https://github.com/user-attachments/assets/04c125d8-eb26-4de6-933d-e7d89c3eaae4">

And here is the main app logic :

<img width="751" alt="Capture d’écran 2024-07-16 à 14 45 39" src="https://github.com/user-attachments/assets/dbe5a861-40b9-440d-a5c9-7703fa610854">

 eveything seems to work fine when loading data and prints successfully the number of objects, but i'm getting 2 different errors,
one error related to the "dataflux_iterable_dataset" class which is the following : 

<img width="776" alt="Capture d’écran 2024-07-16 à 14 48 56" src="https://github.com/user-attachments/assets/8517a7dd-d3c7-4011-9454-9cbf1bbb996f">

i don't know why it's trying to delete the files after the code execution (and prints the number of objects succesfully), but it's not having the permission to delete (NOTE : with the "dataflux_mapstyle_dataset" i do not encounter this issue).

And the other error is related to the "data.DataLoader" from the pytorch module when i try to print the batches (the testing loop i've written above).

<img width="545" alt="Capture d’écran 2024-07-16 à 14 51 14" src="https://github.com/user-attachments/assets/d6fdb9d2-0aff-4238-a4fe-64ddb1dbe205">

Any solutions to this ?

MattIrv:
Thanks for reporting this. 

For the first issue, Dataflux uses a "composed-download" algorithm (documented [here](https://github.com/GoogleCloudPlatform/dataflux-pytorch?tab=readme-ov-file#composite-objects)) for small files where it merges small files together into a larger temporary file, downloads the temporary file, and then deletes the temporary file. It looks like you don't have permission for delete, which is causing problems. If you have a low number of files or performance is not a concern, you can disable this behavior by setting the config as documented in that link. Alternatively, you can make sure you have object delete permission on the bucket you're working with. It's not immediately clear to me why you're not hitting this with the map-style data set as well - maybe I'd need to see a version of your `load_dataset` code to understand better.

I'll also look into whether it's possible for the code to automatically detect that it doesn't have the right permissions for compose/delete and automatically disable that behavior if so.

[...]

PR #67, opened 2024-07-25
For mac and windows skip passing client storage to avoid pickling error in multiprocessing

akansha1812:
Fixes  https://github.com/GoogleCloudPlatform/dataflux-pytorch/issues/58

The pickling error was due to passing storage client to pytorch dataloader. When num_workers>0, pytorch uses multi-processing. Multi-processing fails as it cannot serialize/pickle storage client and lambda function. 

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/dataflux_iterable_dataset.py
+++ b/dataflux_pytorch/dataflux_iterable_dataset.py
@@ -16,8 +16,8 @@
 
 import logging
 import math
+import multiprocessing
 import os
-import platform
 import warnings
 
 import dataflux_core
@@ -29,8 +29,7 @@ from torch.utils import data
 MODIFIED_RETRY = DEFAULT_RETRY.with_deadline(100000.0).with_delay(
     initial=1.0, multiplier=1.5, maximum=30.0)
 
-OS = platform.system()
-LINUX = "Linux"
+FORK = "fork"
 
 
 class Config:
@@ -87,7 +86,7 @@ class Config:
         self.download_retry_config = download_retry_config
 
 
-def data_format_default(self, data):
+def data_format_default(data):
     return data
 
 
@@ -120,9 +119,11 @@ class DataFluxIterableDataset(data.IterableDataset):
                 during initialization.
         """
         super().__init__()
-        if storage_client is not None and OS != LINUX:
+        multiprocessing_start = multiprocessing.get_start_method(
+            allow_none=False)
+        if storage_client is not None and multiprocessing_start != FORK:
             warnings.warn(
-                "Setting the storage client is not fully supported on non-Linux platforms.",
+                "Setting the storage client is not fully supported when multiprocessing starts with spawn or forkserver methods.",
                 UserWarning)
         self.storage_client = storage_client
         self.project_name = project_name
--- a/dataflux_pytorch/dataflux_iterable_dataset.py
+++ b/dataflux_pytorch/dataflux_iterable_dataset.py
@@ -17,6 +17,7 @@
 import logging
 import math
 import os
+import platform
 
 import dataflux_core
 from google.api_core.client_info import ClientInfo
@@ -25,7 +26,11 @@ from google.cloud.storage.retry import DEFAULT_RETRY
 from torch.utils import data
 
 MODIFIED_RETRY = DEFAULT_RETRY.with_deadline(100000.0).with_delay(
-    initial=1.0, multiplier=1.5, maximum=30.0)
+    initial=1.0, multiplier=1.5, maximum=30.0
+)
+
+OS = platform.system()
+LINUX = "Linux"
 
 
 class Config:
@@ -66,10 +71,8 @@ class Config:
         prefix: str = None,
         max_listing_retries: int = 3,
         disable_compose: bool = False,
-        list_retry_config:
-        "google.api_core.retry.retry_unary.Retry" = MODIFIED_RETRY,
-        download_retry_config:
-        "google.api_core.retry.retry_unary.Retry" = MODIFIED_RETRY,
+        list_retry_config: "google.api_core.retry.retry_unary.Retry" = MODIFIED_RETRY,
+        download_retry_config: "google.api_core.retry.retry_unary.Retry" = MODIFIED_RETRY,
     ):
         self.sort_listing_results = sort_listing_results
         self.max_composite_object_size = max_composite_object_size
@@ -89,7 +92,7 @@ class DataFluxIterableDataset(data.IterableDataset):
         project_name,
         bucket_name,
         config=Config(),
-        data_format_fn=lambda data: data,
+        data_format_fn=None,
         storage_client=None,
     ):
         """Initializes the DataFluxIterableDataset.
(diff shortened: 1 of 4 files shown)


dataflux_pytorch/dataflux_iterable_dataset.py
```diff
@@ -111,53 +114,66 @@ def __init__(
                 during initialization.
         """
         super().__init__()
-        self.storage_client = storage_client
-        if not storage_client:
-            self.storage_client = storage.Client(project=project_name, )
-        dataflux_core.user_agent.add_dataflux_user_agent(self.storage_client)
+        self.storage_client = None
+        if OS == LINUX:
```
MattIrv, line 118: It looks like this means we're silently ignoring the passed-in storage client when the user is on other platforms? If so, should we either raise an exception or log a warning when it's set on other platforms?

In fact, it seems like in some cases it may work for the user and for you on Mac, but it's not entirely clear when, is that right? If so we may want to log a warning saying that setting the storage client is not supported on non-Linux platforms but try to do it anyway

dataflux_pytorch/dataflux_iterable_dataset.py
```diff
@@ -111,53 +114,66 @@ def __init__(
...
+                storage_client
+                if storage_client is not None
+                else storage.Client(project=project_name)
+            )
+            dataflux_core.user_agent.add_dataflux_user_agent(self.storage_client)
         self.project_name = project_name
         self.bucket_name = bucket_name
-        self.data_format_fn = data_format_fn
+        self.data_format_fn = (
+            data_format_fn if data_format_fn is not None else self.data_format_default
```
MattIrv, line 128: If the user passes a custom data_format_fn which is a lambda, will it fail to pickle it? If so, should we add a doc comment or something that on non-Linux platforms they need to pass a function definition rather than a lambda?

MattIrv, line 128: I think it may be possible to define data_format_default inline even using "def" syntax rather than maxing it a method on the class

dataflux_pytorch/dataflux_iterable_dataset.py
```diff
@@ -111,53 +114,66 @@ def __init__(
...
-                        dataflux_core.download.dataflux_download_lazy(
-                            project_name=self.project_name,
-                            bucket_name=self.bucket_name,
-                            objects=self.objects,
-                            storage_client=self.storage_client,
-                            dataflux_download_optimization_params=self.
-                            dataflux_download_optimization_params,
-                            retry_config=self.config.download_retry_config,
-                        ))
+            yield from (
```
MattIrv, line 143: Here and elsewhere in this PR, it looks like your formatter might have reformatted some things with different settings than we're using for the project. Can you make sure that the recommended extensions are installed (Extensions > Recommended in VS Code) and then re-run the formatter on these files?

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -115,18 +118,25 @@ def __init__(
             None.
         """
         super().__init__()
-        self.storage_client = storage_client
-        if not storage_client:
-            self.storage_client = storage.Client(project=project_name, )
-        dataflux_core.user_agent.add_dataflux_user_agent(self.storage_client)
+        self.storage_client = None
+        if OS == LINUX:
```
MattIrv, line 122: Same comments on this file as above

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -140,23 +150,26 @@ def __getitem__(self, idx):
...
                 retry_config=self.config.download_retry_config,
             )
         ]
 
+    def data_format_default(self, data):
+        return data
+
     def _list_GCS_blobs_with_retry(self):
         """Retries Dataflux Listing upon exceptions, up to the retries defined in self.config."""
         error = None
```
MattIrv, line 175: There are no tests in this PR, is it possible to add tests that catch these failures, for example by creating dataset objects and trying to pickle them?

dataflux_pytorch/dataflux_iterable_dataset.py
jdnurme: Definitely log a warning at very least to let them know it's being ignored. I would even say throwing an error is probably appropriate so they just remove the setting.

jdnurme: Mac is unstable with fork (see https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) so safest option is to move forward with linux only.

dataflux_pytorch/dataflux_iterable_dataset.py
jdnurme: Lambda is definitely preferable for readability unless you ran into specific errors with passing a lambda.

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -90,7 +93,7 @@ def __init__(
         project_name,
         bucket_name,
         config=Config(),
-        data_format_fn=lambda data: data,
+        data_format_fn=None,
```
jdnurme, line 96: Why is this necessary?

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -140,23 +150,26 @@ def __getitem__(self, idx):
...
                 storage_client=self.storage_client,
-                dataflux_download_optimization_params=self.
-                dataflux_download_optimization_params,
+                dataflux_download_optimization_params=self.dataflux_download_optimization_params,
                 threads=self.config.threads_per_process,
                 retry_config=self.config.download_retry_config,
             )
         ]
 
+    def data_format_default(self, data):
```
jdnurme, line 170: Even if we do need to have this as a separate function, i don't see any reason for it to be a member of the class. Can we define it outside of the class scope?

dataflux_pytorch/dataflux_mapstyle_dataset.py
akansha1812: Multiprocessing in pytorch dataloader cannot pickle lambda functions. 
Getting error: 
"_pickle.PicklingError: Can't pickle <function DataFluxIterableDataset.<lambda> at 0x10a623550>: attribute lookup DataFluxIterableDataset.<lambda> on dataflux_pytorch.dataflux_iterable_dataset failed"

dataflux_pytorch/dataflux_mapstyle_dataset.py
akansha1812: Done

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -284,8 +284,6 @@ def test_init_sets_user_agent(self):
             self.want_objects,
             f"got listed objects {ds.objects}, want {self.want_objects}",
         )
-        self.assertTrue(
-            self.storage_client._connection.user_agent.startswith("dataflux"))
```
MattIrv, line 288: Is the user-agent still getting set somewhere further down the stack?

MattIrv (approved):

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -115,10 +124,11 @@ def __init__(
             None.
         """
         super().__init__()
+        if storage_client is not None and OS != LINUX:
```
MattIrv, line 127: For these checks, rather than checking the OS, would it make more sense to check the set multiprocessing start_method instead? I think that would be more reliable, since even Linux users might use "spawn" in some cases https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_start_method

MattIrv (approved):

MattIrv (approved):

PR #68, opened 2024-07-26
Update demo loops to configure multiprocessing start method

MattIrv:
This makes it easy to reproduce and test the issues in https://github.com/GoogleCloudPlatform/dataflux-pytorch/issues/58. 

I confirmed that the following fails right now:

- "spawn", with or without an initial client set
- "forkserver", with or without an initial client set

When I pulled in Akansha's PR https://github.com/GoogleCloudPlatform/dataflux-pytorch/pull/67, the above scenarios only fail when I set the initial client.

I also had to pull the `read_image_modified` definition out of inline since it couldn't be picked in that case.

- [x] Tests pass

### DIFF
--- a/demo/list-and-download/iterable/simple_iterable_dataset.py
+++ b/demo/list-and-download/iterable/simple_iterable_dataset.py
@@ -16,8 +16,10 @@
 
 import argparse
 import logging
+import multiprocessing
 import time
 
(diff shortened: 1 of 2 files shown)


akansha1812 (approved):
Thanks Matt!

PR #69, opened 2024-07-29
Added lightning package to setup file and updated version for re…

divrawal:
new release with updated lightning dependency

- [x] Tests pass

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,6 @@ dependencies = [
 setup(
     name="dataflux-pytorch",
     packages=["dataflux_core", "dataflux_pytorch", "dataflux_pytorch.lightning"],
-    package_dir={"dataflux_core": "./dataflux_client_python"},
+    package_dir={"dataflux_core": "dataflux_client_python/dataflux_core"},
     install_requires=dependencies,
 )
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-from setuptools import setup, find_packages
+from setuptools import setup
 
 dependencies = [
     "torch",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -10,7 +10,7 @@
 ]
 setup(
     name="dataflux-pytorch",
-    packages=["dataflux_core", "dataflux_pytorch"],
+    packages=["dataflux_core", "dataflux_pytorch", "dataflux_pytorch.lightning"],
```
bernardhan33, line 13: Is this line causing the problem? Ideally we should be using `find_packages()` ([ref](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#packages)) to find the packages so that we don't need to specify this every time a new package is added (error-prone).

We probably took a short path because we needed "dataflux_core" to map to "dataflux_client_python/dataflux_core" (see below). Ideally we should be just importing "dataflux_client_python.dataflux_core" instead of "dataflux_core". Could you investigate if that path is feasible?

setup.py
```diff
@@ -10,7 +10,6 @@
 ]
 setup(
     name="dataflux-pytorch",
-    packages=["dataflux_core", "dataflux_pytorch"],
-    package_dir={"dataflux_core": "dataflux_client_python/dataflux_core"},
+    packages=find_packages(exclude=["*tests*"]),
```
bernardhan33, line 13: Yeah this looks promising until http://sponge2.corp.google.com/62d95cec-1727-41cc-9368-4a9e84bd2ee6 and you can see "E   ModuleNotFoundError: No module named 'dataflux_core'". Tricky business isn't it.

MattIrv:
@divrawal can you update the PR title/description to describe what you're trying to accomplish with this PR?

setup.py
bernardhan33: I see we are reverting to not using "find_packages()", which is fine. But could you file a follow-up ticket to investigate the solutions here? Ideally we would not want every extra directory (package) to be declared every time they are added?

bernardhan33:
Approving this in light of the follow-up task b/356629540. Divya is adding all the details to that ticket. In short we will let this one pass to unblock the other work. But we should prioritize b/356629540 because similar changes here are risky -- we should rely on `find_packages()` to automatically find all the packages to publish and fix the import paths of "dataflux_core" to "dataflux_client_python.dataflux_core".

bernardhan33 (approved):

MattIrv:
@divrawal is this ready to merge?

PR #71, opened 2024-08-01
continuous test which installs dataflux from PyPi and runs the unit

akansha1812:
Fixes #<issue_number_goes_here>

- [ ] Tests pass

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,34 @@
+name: Continuous Tests
+
+on:
+  push:
+    branches: [ amaloo/pypi-tests ]
+  pull_request:
+    branches: [ amaloo/pypi-tests ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.8' 
+
+     - name: Create virtual environment
+      run: python3 -m venv venv
+
+    - name: Activate virtual environment
+      shell: bash
+      run: source venv/bin/activate 
+
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade pip
+
+        pip install gcs-torch-dataflux
+    - name: Run tests
+      run: |
+        pytest dataflux_pytorch/tests -vv --junit-xml=sponge_log.xml --log-cli-level=DEBUG


kokoro/pypi.sh
```diff
@@ -0,0 +1,49 @@
...
+
+    echo Installing required dependencies.
+    pip install .
+    python3 -m pip install --upgrade pip
+    pip install gcs-torch-dataflux
+}
+
+function run_unit_tests() {
+    echo Running unit tests.
+    python3 -m pytest dataflux_pytorch/tests -vv --junit-xml="${KOKORO_ARTIFACTS_DIR}/unit_tests/sponge_log.xml" --log-cli-level=DEBUG
```
MattIrv, line 44: Have you been able to confirm somehow that this actually runs the tests against the version that is installed from PyPi? Since it looks like the exact same command that the regular configuration uses to run the tests, I'm guessing it might not.

akansha1812: I'm trying to figure that out. Please ignore this PR.

akansha1812: Changed command by removing `python3 -m`.

The integration test is calling pypi version of  `dataflux_pytorch` installed in `test_env/lib/python3.11/site-packages`.

https://screenshot.googleplex.com/ArZKEYfffZ7GZuN.png

kokoro/pypi.sh
```diff
@@ -0,0 +1,59 @@
...
+    echo Installing python3-pip.
+    sudo apt-get -y install python3-pip
+
+    echo Installing project from pypi
+    pip install gcs-torch-dataflux
+}
+
+function run_integration_tests(){
+    echo Running integration tests.
+    pytest dataflux_pytorch/integration_tests/integration_test.py -vv --junit-xml="${KOKORO_ARTIFACTS_DIR}/integration_tests/sponge_log.xml" --log-cli-level=DEBUG
```
bernardhan33, line 53: What about also running the unit tests?

kokoro/pypi.sh
```diff
@@ -0,0 +1,59 @@
...
+
+function install_requirements() {
+    echo Installing python3-pip.
+    sudo apt-get -y install python3-pip
+
+    echo Installing project from pypi
+    pip install gcs-torch-dataflux
+}
+
+function run_integration_tests(){
```
MattIrv, line 51: Could you add a comment here clarifying that the intention is to run the tests from PyPi? That way someone doesn't come along later and accidentally update it to run the local version

MattIrv (approved):
Great work figuring this out! It will be really good to have these tests running so we don't accidentally break things

kokoro/pypi.sh
akansha1812: Currently the versions in pypi and in github are different. Using pypi version for unit test can fail frequently.

bernardhan33: Could you clarify? `Using pypi version for unit test can fail frequently.` to me sounds like if I'm a user who installs gcs-torch-dataflux from PyPI and run unit tests locally, the unit tests can fail? If so we need to debug what's going wrong and it's an indicator that we will need to include this path in the continuous tests.

akansha1812: PyPi version is few changes behind github main branch. 
In the main branch, I can see a `/dataflux-pytorch/dataflux_pytorch/_helper.py` , while this does not exists in pypi version.

bernardhan33: oh sorry I misunderstood. yeah you are right. When you install the PyPI version you don't have "unit tests" in it as we didn't package it in the release, and we shouldn't do either.

So we only need to run the integration tests with the installed PyPI version, that's right.

PR #73, opened 2024-08-02
Lightning multinode parquet

jdnurme:
Functional single and distributed examples of dataflux used for parquet-text file processing with lightning. README will follow this PR with more detailed instructions on how to use and run this demo.

- [x] Tests pass


demo/lightning/text-based/distributed/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Base Image: Start with a slim Python image that already includes Python3.
```
MattIrv, line 1: In order to simplify things, you might want to make this work with the Dockerfile in the base of the repo. That also has the benefit of packaging up the whole repo including local changes, so that you can make and test changes without pushing them to PyPi

demo/lightning/text-based/distributed/model.py
```diff
@@ -0,0 +1,309 @@
...
+    config = df_iter.Config(prefix=args.prefix, sort_listing_results=True)
+    my_model = TextDemoModel(encoder, decoder)
+    bsize = args.batch_size
+
+    # Construct the lightning trainer and run the fit with our model and custom dataset.
+    # Note that limit_train_batches specifies how much of the dataset to train each epoch.
+    trainer = pl.Trainer(accelerator='cpu',
+                         devices=args.devices,
+                         strategy=pl.strategies.DDPStrategy(
+                             timeout=datetime.timedelta(minutes=30)),
```
MattIrv, line 276: Can you update this to a smaller timeout here so that it fails faster if something is wrong and doesn't appear to be stuck?

demo/lightning/text-based/distributed/requirements.txt
```diff
@@ -0,0 +1,6 @@
+torch
+dataflux
```
MattIrv, line 2: Is this supposed to be here? I think gcs-torch-dataflux is the right one isn't it?

demo/lightning/text-based/single-node/requirements.txt
```diff
@@ -0,0 +1,6 @@
+torch
+dataflux
```
MattIrv, line 2: Same comment here

demo/lightning/text-based/distributed/Dockerfile
MattIrv, line 1: Either in this PR or a follow up, can you add a README?

MattIrv (approved):

demo/lightning/text-based/distributed/example-deploy.yaml
```diff
@@ -0,0 +1,72 @@
+# Use Static provisioning to define a PV for Cloud Storage buckets.
```
bernardhan33, line 1: Remove this comment too as we are not using GCSFuse.

demo/lightning/text-based/distributed/example-deploy.yaml
```diff
@@ -0,0 +1,72 @@
...
+# https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/cloud-storage-fuse-csi-driver#provision-static.
+apiVersion: jobset.x-k8s.io/v1alpha2
```
bernardhan33, line 3: Note that the YAML file here assumes that you have jobset and Kueue enabled so it will not work with any regular GKE cluster. I see that you have a README PR following up. We might as well ask that users create a cluster using XPK which comes with those things. In fact you are able to run this on our clusters because those are created by XPK.

demo/lightning/text-based/distributed/example-deploy.yaml
```diff
@@ -0,0 +1,72 @@
...
+      template:
+        spec:
+          parallelism: 2    # Equal to the number of VMs per slice
+          completions: 2    # Same as the above.
+          backoffLimit: 0   # When any pod fails, the job is failed
+          template:
+            spec:
+              schedulerName: default-scheduler
+              restartPolicy: Never
+              nodeSelector:
```
bernardhan33, line 26: remove these too because it is not guaranteed for clusters to have local SSDs.

demo/lightning/text-based/distributed/Dockerfile
bernardhan33, line 1: For all the files here, please add the header license.

demo/lightning/text-based/distributed/requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+pytorch-lightning
```
bernardhan33, line 3: gcs-torch-dataflux should contain lightning?

demo/lightning/text-based/distributed/Dockerfile
jdnurme: Yes that will be added in the followup

jdnurme: Yes i'll update that alongside the readme that includes instructions on how to run

PR #74, opened 2024-08-02
Make DatafluxPytTrain a wrapper of DataFluxMapStyleDataset

abhibyreddi:
Also removes Ray specific code as we figured out how to deploy on GKE without relying on Ray.

Ran `train.py` locally and made sure it ran successfully.


MattIrv (approved):
Nice way of simplifying the code!

demo/lightning/image-segmentation/dataset.py
```diff
@@ -31,99 +30,68 @@ def __init__(
...
-        self.labels = dataflux_core.fast_list.ListingController(
-            max_parallelism=self.config.num_processes,
-            project=self.project_name,
-            bucket=self.bucket_name,
-            sort_results=self.config.
-            sort_listing_results,  # This needs to be True to map images with labels.
-            prefix=labels_prefix,
-        ).run()
+
+        self.images_in_bytes = dataflux_mapstyle_dataset.DataFluxMapStyleDataset(
```
bernardhan33, line 47: nit: might be better named with something like "images_dataset" because we used "images_in_bytes" to represent the downloaded images files in the format of bytes. Here we are just defining a dataset and haven't downloaded anything yet.

demo/lightning/image-segmentation/dataset.py
```diff
@@ -31,99 +30,68 @@ def __init__(
...
+            project_name=self.project_name,
+            bucket_name=self.bucket_name,
+            config=dataflux_mapstyle_dataset.Config(
+                sort_listing_results=True,
+                prefix=kwargs["images_prefix"],
+            ),
+        )
+
+        self.labels_in_bytes = dataflux_mapstyle_dataset.DataFluxMapStyleDataset(
```
bernardhan33, line 56: ditto

demo/lightning/image-segmentation/dataset.py
```diff
@@ -31,99 +30,68 @@ def __init__(
...
-            sort_results=self.config.
-            sort_listing_results,  # This needs to be True to map images with labels.
-            prefix=labels_prefix,
-        ).run()
+
+        self.images_in_bytes = dataflux_mapstyle_dataset.DataFluxMapStyleDataset(
+            project_name=self.project_name,
+            bucket_name=self.bucket_name,
+            config=dataflux_mapstyle_dataset.Config(
+                sort_listing_results=True,
```
bernardhan33, line 51: nit: would be good to keep the comment here that "# This needs to be True to map images with labels.". This needs to be set for the two datasets to map data samples to each other.

demo/lightning/image-segmentation/dataset.py
```diff
@@ -31,99 +30,68 @@ def __init__(
...
 
     def __getitem__(self, idx):
         image = np.load(
             io.BytesIO(
-                dataflux_core.download.download_single(
-                    storage_client=self.storage_client,
-                    bucket_name=self.bucket_name,
-                    object_name=self.images[idx][0],
-                )), )
+                self.images_in_bytes.objects[idx],
```
bernardhan33, line 71: please fix: This will break the __getitem__ logic as the "objects" variable in the dataset just refers to the listing results. Instead, you will need `self.images_in_bytes[idx]` where `images_in_bytes` is the dataset.

demo/lightning/image-segmentation/dataset.py
```diff
@@ -31,99 +30,68 @@ def __init__(
...
+        )
 
         label = np.load(
             io.BytesIO(
-                dataflux_core.download.download_single(
-                    storage_client=self.storage_client,
-                    bucket_name=self.bucket_name,
-                    object_name=self.labels[idx][0],
-                )), )
+                self.labels_in_bytes.objects[idx],
```
bernardhan33, line 77: ditto

demo/lightning/image-segmentation/dataset.py
abhibyreddi: Thanks for catching! Sent out PR #77 to address all comments on this PR. Thanks for reviewing on you day off.

bernardhan33: Thanks for the quick fix. I scheduled a /sendLater for Monday morning but you beat me to it ;)

Signed off early on Friday so making the time up for review.

PR #81, opened 2024-08-06
Add code to make it possible to deploy training on a multi-node GKE cluster

abhibyreddi:
README file will be added in the follow up PR

Tested by deploying on four nodes


MattIrv (approved):
Just to confirm, does this code still work for single-node local runs as well?

abhibyreddi:
> Just to confirm, does this code still work for single-node local runs as well?

Added a new flag called "local" which defaults to False. It's now possible to run it locally on a single-node machine.

PR #82, opened 2024-08-07
Configure shared memory size

abhibyreddi:
Default limit is too low (64MiB) causing OOMs for all cases where `num_workers > 0`. 

Tested by running successfully on 4 nodes with `prefetch_factor` set to `100` and `num_workers` set to `1`.

### DIFF
--- a/demo/lightning/image-segmentation/data.py
+++ b/demo/lightning/image-segmentation/data.py
@@ -65,7 +65,7 @@ class Unet3DDataModule(pl.LightningDataModule):
             num_workers=self.args.num_workers,
             pin_memory=True,
             drop_last=True,
-            prefetch_factor=100,
+            prefetch_factor=self.args.prefetch_factor,
         )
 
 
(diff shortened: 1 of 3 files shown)


demo/lightning/image-segmentation/data.py
```diff
@@ -62,9 +62,10 @@ def train_dataloader(self):
             batch_size=self.args.batch_size,
             shuffle=not self.args.benchmark and self.train_sampler is None,
             sampler=self.train_sampler,
-            num_workers=0,
+            num_workers=self.args.num_workers,
             pin_memory=True,
             drop_last=True,
+            prefetch_factor=100,
```
MattIrv, line 68: It would probably be a good idea to make this configurable with a flag too

MattIrv (approved):
Nice fix! Do you see an improvement in performance when running with more workers now?

demo/lightning/image-segmentation/data.py
abhibyreddi: Done!

PR #87, opened 2024-08-08
Initialize new storage_client.bucket on every request

Yash9060:
Remove the check to initialize bucket client only once, since we want to be able to pass different GCS buckets in save/restore/delete checkpointing functions.

This is in regards to comment on https://github.com/GoogleCloudPlatform/dataflux-pytorch/pull/85

- [X] Tests pass

### DIFF
--- a/dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
+++ b/dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, Tuple
 
 import torch
 from dataflux_core import user_agent
@@ -17,7 +17,6 @@ class DatafluxLightningCheckpoint(CheckpointIO):
     ):
         self.project_name = project_name
         self.storage_client = storage_client
-        self.bucket = None
         if not storage_client:
             self.storage_client = storage.Client(project=self.project_name, )
         user_agent.add_dataflux_user_agent(self.storage_client)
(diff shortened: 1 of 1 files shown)


dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
```diff
@@ -52,8 +52,7 @@ def _parse_gcs_path(self, path: Union[str, Path]) -> str:
             bucket_name, prefix = split
         if not bucket_name:
             raise ValueError("Bucket name must be non-empty")
-        if not self.bucket:
-            self.bucket = self.storage_client.bucket(bucket_name)
+        self.bucket = self.storage_client.bucket(bucket_name)
```
MattIrv, line 55: I wonder if this could cause problems in any sort of multi-threaded scenario, if one thread calls _parse_gcs_path, then another thread calls _parse_gcs_path, then the original thread makes its actual API call.

To avoid any problems like that, would it work to initialize it wherever self.bucket is used today?

Yash9060: As discussed offline initializing bucket client where we need it.

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
```diff
@@ -52,18 +51,17 @@ def _parse_gcs_path(self, path: Union[str, Path]) -> str:
             bucket_name, prefix = split
         if not bucket_name:
             raise ValueError("Bucket name must be non-empty")
-        if not self.bucket:
-            self.bucket = self.storage_client.bucket(bucket_name)
-        return prefix
+        return prefix, bucket_name
```
MattIrv, line 54: nit: It might be more intuitive to return these in the reverse order, since the bucket name comes before the prefix in the original string

MattIrv (approved):

dataflux_pytorch/lightning/dataflux_lightning_checkpoint.py
Yash9060: Agreed. Changed the order.

PR #88, opened 2024-08-08
Submitting example ray code that runs distributed checkpointing 

divrawal:
Displays how to configure dataflux checkpointing and can be used to benchmark checkpoints.

- [ ] Tests pass

### DIFF
--- a/demo/lightning/checkpoint/ray/README.md
+++ b/demo/lightning/checkpoint/ray/README.md
@@ -1,0 +1,69 @@
+## Deployment
+
+### Cluster
+
+We use [Ray.IO](Ray.IO) to deploy and scale the training workload with [PyTorch](https://pytorch.org/) as the underlying machine learning framework.
+
+We will spin up a [Ray cluster](https://docs.ray.io/en/latest/cluster/key-concepts.html#ray-cluster) for our workload. The [cluster.yaml](cluster.yaml) file contains the specification and here's the overview:
+
+
+
+*   Ray head node
+    *   There’s only one head node in the cluster that is responsible for cluster management. See more about the Ray head node [here](https://docs.ray.io/en/latest/cluster/key-concepts.html#head-node).
+    *   Key Specification:
+        *   **Machine type:** n1-standard-32
+        *   **Source Image:** projects/deeplearning-platform-release/global/images/family/common-cpu
+*   Ray worker node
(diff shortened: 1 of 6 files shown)


demo/lightning/checkpoint/ray/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+2. Submit the workload job by running
+
+```shell
+export RAY_ADDRESS=http://localhost:8265
+python3 submit.py
+```
+
+## Results
+The job should finish within a couple minutes and you can follow the job log by clicking into the job detail from the Ray dashboard. Time to save each distributed checkpoint will be printed out to logs and can be viewed on job completion.
```
MattIrv, line 66: Can you point me to logs or output from a successful run? What cluster are you using?

demo/lightning/checkpoint/ray/train.py
```diff
@@ -0,0 +1,143 @@
...
+    train_dataloader = train_ds.iter_torch_batches(batch_size=batch_size_per_worker)
+
+    trainer = Trainer(
+        default_root_dir=config["flags"].default_root_dir,
+        max_epochs=config["flags"].epochs,
+        devices="auto",
+        max_steps=1,
+        accelerator="auto",
+        strategy=strategy,
+        plugins=[RayLightningEnvironment(),ckpt],
```
MattIrv, line 92: When I remove ckpt here it doesn't seem to make any difference in behavior, and my checkpoint still shows up in GCS. I also see from the metrics that the writes to my bucket don't have the "dataflux" user-agent. This makes me think that Ray is probably checkpointing via gcsfs by default, and ignoring the checkpointer you passed in here. Were you able to confirm that your checkpoint code was called at all? If not we can still check this in as a good example of how to configure Ray to checkpoint for GCS (and potentially get baseline metrics from it), but I don't think it really uses Dataflux.

MattIrv (approved):

demo/lightning/checkpoint/ray/train.py
divrawal: I did confirm the dataflux checkpointing code was called because without configuration it results in error but Ray still may be able to overwrite some behavior. Let me dig into the stack trace more

dataflux_client_python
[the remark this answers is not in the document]
divrawal: This is not intentional. I think I need to update my submodule

PR #92, opened 2024-08-15
Skip training when demo is run to benchmark Dataflux

abhibyreddi:
Adds an option to run the demo to benchmark Dataflux.

- [X] Tests pass; ran locally


demo/lightning/image-segmentation/unet3d.py
```diff
@@ -29,7 +29,7 @@ def __init__(self,
                  weights_init_scale=1.0):
         super(Unet3D, self).__init__()
 
-        filters = [32, 64, 128, 256, 320]
+        filters = [2, 2, 2, 2, 2]
```
MattIrv, line 32: Do we still want to update this even for the demo version?

MattIrv (approved):

demo/lightning/image-segmentation/unet3d.py
abhibyreddi: Thanks for catching. Updated so that filters gets set to smaller values only when run in benchmark mode.

PR #94, opened 2024-08-19
Raise exception when ADC are missing

abhibyreddi:
Dataflux throws NameError when there is an authentication error: http://shortn/_6aUnC3U0bl
After this change: http://shortn/_H8ceJgcMzC

Also removed unused imports from `dataflux_iterable_dataset.py` and `dataflux_mapstyle_dataset.py`

- [X] Tests pass; ran demo code locally and verified that this change works


dataflux_pytorch/_helper.py
```diff
@@ -32,5 +32,6 @@ def _get_missing_permissions(storage_client: any, bucket_name: str,
         perm = bucket.test_iam_permissions(required_perm)
     except Exception as e:
         logging.exception(f"Error testing permissions: {e}")
+        return None
```
MattIrv, line 35: In the screenshots you shared, you can see that it initially logs this exception, but then fails with a different error, "[rank0]: RuntimeError: multiprocessing child process became unresponsive; check logs for underlying error".

Rather than returning "None" hear, which seems to log this exception but allow things to continue, should we re-raise this exception here so that Dataflux exits at this point when it can't test the permissions?

abhibyreddi: Done! Stack trace now looks like http://shortn/_HtXlitW04U

MattIrv: Thanks! You can see now that this results in the same error getting logged twice. It may be better to just allow the exception to get raised and handled/logged by the caller, rather than also logging it here

MattIrv (approved):

dataflux_pytorch/_helper.py
abhibyreddi: Thanks for the quick feedback Matt! Changed it a bit. This is what uses would see now if default creds have expired or are missing: http://shortn/_utwTn2q1mO

Could you take a look at the additional changes I made to the other two files?

dataflux_pytorch/_helper.py
```diff
@@ -30,7 +31,7 @@ def _get_missing_permissions(storage_client: any, bucket_name: str,
 
     try:
         perm = bucket.test_iam_permissions(required_perm)
-    except Exception as e:
-        logging.exception(f"Error testing permissions: {e}")
+    except Exception:
+        raise
```
MattIrv, line 35: No need to have the try/except block here now, it's just the same as doing this outside of the block and letting the exception raise itself

dataflux_pytorch/dataflux_iterable_dataset.py
```diff
@@ -136,11 +135,18 @@ def __init__(
...
-                required_perm=[CREATE, DELETE])
+            try:
+                missing_perm = _get_missing_permissions(
+                    storage_client=self.storage_client,
+                    bucket_name=self.bucket_name,
+                    project_name=self.project_name,
+                    required_perm=[CREATE, DELETE])
+            except RefreshError as e:
+                e.add_note(
+                    "Default credentials might be missing. Run `gcloud auth application-default login`")
```
MattIrv, line 146: Maybe this should point to the GCP docs page on configuring app-default credentials instead? Running the gcloud command would be the right advice when running locally, but won't help users who are on GKE for example.

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -140,11 +139,18 @@ def __init__(
...
-                required_perm=[CREATE, DELETE])
+            try:
+                missing_perm = _get_missing_permissions(
+                    storage_client=self.storage_client,
+                    bucket_name=self.bucket_name,
+                    project_name=self.project_name,
+                    required_perm=[CREATE, DELETE])
+            except RefreshError as e:
+                e.add_note(
+                    "Default credentials might be missing. Run `gcloud auth application-default login`")
```
MattIrv, line 150: Same comment here as above.

Also if we are going to add this note in everywhere it occurs, why not add it just in _helper.py instead of both dataflux_iterable_dataset.py and dataflux_mapstyle_dataset.py?

dataflux_pytorch/_helper.py
abhibyreddi: Reverted the changes I made to the other two files and updated the exception message here. Stack trace now looks like this: http://shortn/_T9BGgSybjP

PR #95, opened 2024-08-19
Allow running demo code in listing-only mode

abhibyreddi:
Training in skipped when `--listing_only` is passed to `train.py` and exits right after initializing the map style datasets.

- [X] Tests pass; ran locally and verified listing time is output when this flag is set

### DIFF
--- a/demo/lightning/image-segmentation/README.md
+++ b/demo/lightning/image-segmentation/README.md
@@ -3,7 +3,7 @@ The code examples in this directory demonstrate how GCS Connector for Pytorch ca
 
 If you are looking to benchmark data loading performance, pass `--benchmark` to `train.py`. Note that this will run the training loop but does not run the actual training logic.
 
-If you are looking to benchmark just the listing performance, pass `--listing_only` to `train.py`. Note that this will skip training altogether.
+If you are looking to benchmark the listing performance only, pass `--listing_only` to `train.py`. Note that this will skip training altogether. Passing `--benchmark` along with `--listing_only` would have the same effect as passing just the latter.
 
 
 1. `dataset.py`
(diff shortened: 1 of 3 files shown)


demo/lightning/image-segmentation/README.md
```diff
@@ -1,7 +1,10 @@
 # Image Segmentation Demo Code
 The code examples in this directory demonstrate how GCS Connector for Pytorch can be used for image segmentation training alongside PyTorch Lightning. The image segmentation workload implemented here works on the [KiTS19](https://github.com/neheller/kits19) dataset which contains `210` images and their corresponding labels. The images and their labels are stored in separate directories.
 
-If you are looking to benchmark data loading performance, pass `--benchmark` to `train.py`. Note that this will skip training altogether.
+If you are looking to benchmark data loading performance, pass `--benchmark` to `train.py`. Note that this will run the training loop but does not run the actual training logic.
+
+If you are looking to benchmark just the listing performance, pass `--listing_only` to `train.py`. Note that this will skip training altogether.
```
MattIrv, line 6: Does this also need the `--benchmark` flag? Does it make a difference?

demo/lightning/image-segmentation/train.py
```diff
@@ -51,8 +51,15 @@ def init_processes():
     if flags.benchmark:
         profiler = "simple"
 
-    model = Unet3DLightning(flags)
+    listing_start = time.time()
     train_data_loader = Unet3DDataModule(flags)
+    listing_end = time.time()
+    if flags.listing_only:
+        print(
```
MattIrv, line 58: nit: Should we print the list time regardless?

MattIrv (approved):

demo/lightning/image-segmentation/train.py
abhibyreddi: Added "Passing `--benchmark` along with `--listing_only` would have the same effect as passing just the latter."

demo/lightning/image-segmentation/README.md
abhibyreddi: Passing `--benchmark` along with `--listing_only` would have the same effect as passing just the latter.

demo/lightning/image-segmentation/train.py
abhibyreddi: Updated so that list time is printed regardless of the mode. The user is notified that training is being skipped when `--listing_only` is passed: http://shortn/_wTll5QX02T

PR #96, opened 2024-08-20
Update defaults for batch_size and num_workers

abhibyreddi:
Update command in deployment file with batch_size and num_workers set to values that gave us best egress BW for 150MB dataset

Also, `Unet3DLightning.training_step` now returns a scalar (that represents loss) instead of None to skip actual training. This is because pytorch lightning does not allow skipping training when run on multiple nodes (returning None on single node training works just fine!)

- [X] Tests pass; deployed on a 10 node cluster and verified that the workload finished successfully

### DIFF
--- a/demo/lightning/image-segmentation/model.py
+++ b/demo/lightning/image-segmentation/model.py
@@ -56,7 +56,7 @@ class Unet3DLightning(pl.LightningModule):
     def training_step(self, train_batch, batch_idx):
         images, labels = train_batch
         if self.benchmark:
-            return None
+            return torch.tensor(1, dtype=float, requires_grad=True)
         predictions = self.model(images)
         loss = self.loss_fn(predictions, labels)
         return loss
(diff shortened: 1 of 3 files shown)


demo/lightning/image-segmentation/model.py
```diff
@@ -56,7 +56,8 @@ def configure_optimizers(self):
     def training_step(self, train_batch, batch_idx):
         images, labels = train_batch
         if self.benchmark:
-            return None
+            # Return a scalar of value 1 to speed up training_step.
```
MattIrv, line 59: nit: This comment doesn't really explain why this speeds it up compared to returning None. Is there any way to add that, either to the comment or to the PR description?

MattIrv (approved):

demo/lightning/image-segmentation/model.py
abhibyreddi: Updated the comment just now :)

abhibyreddi: Also updated PR description.

PR #105, opened 2024-09-09
Adding GCS Custom reader

Yash9060:
This adds custom GCS Reader to dataflux pytroch. 

Demo  POC is present in yashsha-tp branch (https://github.com/GoogleCloudPlatform/dataflux-pytorch/compare/yashsha-tp?expand=1) . I will add it to demo section after this PR is merged. 

create_stream is called with either "wb" (write binary) https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/filesystem.py#L318 

or "rb"(read binary) mode https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/filesystem.py#L643 & https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/filesystem.py#L679 

In read mode we need to pass Bytes instead of stream of bytes otherwise pickle.load fails. 

- [X] Tests pass

### DIFF
--- a/dataflux_pytorch/lightning/gcs_filesystem.py
+++ b/dataflux_pytorch/lightning/gcs_filesystem.py
@@ -6,7 +6,7 @@ from typing import Generator, Optional, Union, cast
 
 from dataflux_core import user_agent
 from google.cloud import storage
-from torch.distributed.checkpoint import FileSystemWriter
+from torch.distributed.checkpoint import FileSystemWriter, FileSystemReader
 from torch.distributed.checkpoint.filesystem import FileSystemBase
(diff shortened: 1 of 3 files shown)


dataflux_pytorch/lightning/gcs_filesystem.py
MattIrv, line 1: Are there tests that need to be updated or added?

MattIrv (approved):

dataflux_pytorch/lightning/gcs_filesystem.py
Yash9060: Added some tests for invalid mode. (although I dont think they add any value here)

PR #106, opened 2024-09-10
MultiNode demo

Yash9060:
Internal Issue: 358678939
Divided checkpointing demo directory into singlenode and multinode with each having their own README.md and yaml deployment file.

Success Logs added to above mentioned bug ID

- [X] Tests pass


demo/lightning/checkpoint/multinode/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Running locally
+
+1. Set the environment variables required to run the demo. These include:
+  * `PROJECT`: The GCP project you are using
+  * `CKPT_DIR_PATH`: The full path of the directory in which to save checkpoints, in the format `gs://<bucket>/<directory>/`
+2. Set the optional environment variables, if desired:
+  * `NUM_LAYERS`: The number of layers in the model, which affects the size of the model and therefore the size of the checkpoints
+  * `ACCELERATOR`: Set to `gpu` if running on a GPU, or `cpu` if running on a CPU (default)
+    * If running on a GPU, you also must set `PJRT_DEVICE` to `CUDA`.
```
MattIrv, line 17: I thought you were having problems with this. Should it say CPU instead?

demo/lightning/checkpoint/multinode/README.md
```diff
@@ -0,0 +1,40 @@
...
+  * `TRAIN_STRATEGY`: This demo requires custom strategy which should override DDP Strategy or FSDP Strategy. FSDP strategy will require actual GPU.
```
MattIrv, line 18: What are the valid values for this environment variable?

demo/lightning/checkpoint/multinode/example-deploy.yaml
```diff
@@ -0,0 +1,103 @@
...
+                
+              priorityClassName: medium
+              hostNetwork: true
+              dnsPolicy: ClusterFirstWithHostNet
+              terminationGracePeriodSeconds: 30
+              containers:
+              - name: checkpoint-gpu
+                image: gcr.io/my-project/my-container
+
+                resources:
```
MattIrv, line 47: Add a comment to remove this if running the CPU-based demo. Or it may even be a good idea to provide 2 example yaml files, one for CPU and one for GPU

demo/lightning/checkpoint/multinode/train.py
```diff
@@ -0,0 +1,236 @@
...
+                      max_steps=3,
+                      accelerator=accelerator,
+                      strategy=DatafluxFSDPStrategy(
+                          path=new_path,
+                          project_name=project,
+                          storage_client=None,
+                          model=model,
+                          state_dict_type="sharded",
+                      ),
+                      num_nodes=1
```
MattIrv, line 214: Is this right, given that this is a multi-node demo?

demo/lightning/checkpoint/multinode/train.py
```diff
@@ -0,0 +1,236 @@
...
+                      )
+    trainer.fit(model, dataloader)
+    new_path = ckpt_dir_path + \
+        "lightning_logs/version_0/checkpoints/checkpoint-epoch=00-step=03.ckpt"
```
MattIrv, line 196: Hardcoded path needs to be removed

demo/lightning/checkpoint/multinode/train.py
```diff
@@ -0,0 +1,236 @@
...
+    print("Restoring checkpoints ...")
+    model = DemoTransformer(vocab_size=dataset.vocab_size,
+                            nlayers=int(os.environ.get("NUM_LAYERS", 2)))
+    trainer = Trainer(default_root_dir=ckpt_dir_path,
+                      plugins=[],
+                      callbacks=[],
+                      min_epochs=4,
+                      max_epochs=5,
+                      max_steps=3,
```
MattIrv, line 205: These should probably all be configurable

demo/lightning/checkpoint/multinode/train.py
```diff
@@ -0,0 +1,236 @@
...
+from dataflux_core import user_agent
+from dataflux_pytorch.lightning import DatafluxLightningCheckpoint
+from lightning.pytorch.trainer.states import TrainerFn
+from lightning.pytorch.strategies import FSDPStrategy
+from lightning.pytorch.strategies.fsdp import _METADATA_FILENAME
+from torch.distributed.checkpoint import save, load
+from torch.nn import Module
+
+
+class DatafluxFSDPStrategy(FSDPStrategy):
```
MattIrv, line 28: Is the long-term plan for this code to leave it as a demo or actually make it a part of the Dataflux library code eventually?

demo/lightning/checkpoint/multinode/README.md
Yash9060: Yes changed it to CPU although It just doesnt feel right that, setting it to CUDA is not an issue for single node but errors out on multi node.

demo/lightning/checkpoint/multinode/train.py
Yash9060: Done replaced it by os environment.

demo/lightning/checkpoint/multinode/train.py
Yash9060: That is a good question, although answer to this really depends on the customer use case. My though process is, if the customers are mostly going to use either FSDP or DDP strategy then we can provide custom classes which extends FSDP/DDP. 

This is something that needs some discussion, but for now I plan to keep this here in the demo, so that it can be added to continuous testing.

demo/lightning/checkpoint/multinode/README.md
Yash9060: Yes changed it to CPU although It just doesnt feel right that, setting it to CUDA is not an issue for single node but errors out on multi node.

demo/lightning/checkpoint/multinode/example-deploy.yaml
Yash9060: ACK will add CPU based demo in follow up PR (which will extend DDP strategy)

demo/lightning/checkpoint/multinode/README.md
```diff
@@ -0,0 +1,40 @@
+# Checkpoint Demo for PyTorch Lightning
+
+The code in this folder provides a training demo for multi node checkpointing with PyTorch Lightning. This demo is under development.
+
+## Limitations
+* Custom Strategy needs to be implemented and passed to the trainer.
```
jdnurme, line 6: We should probably hyperlink to pytorch strategy docs here (i think this is it but please double check https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.strategies.Strategy.html)

demo/lightning/checkpoint/multinode/README.md
```diff
@@ -0,0 +1,40 @@
...
+* `requirements.txt` includes gcsfs because even though it is not used for checkpointing, PyTorch Lightning's default logger also writes to the root directory where checkpoints are saved.
+
+## Running locally
+
+1. Set the environment variables required to run the demo. These include:
+  * `PROJECT`: The GCP project you are using
```
jdnurme, line 12: Not strictly necessary but we pass these in as command line args in the other demos. Generally this allows for all info to be known from the python call.

demo/lightning/checkpoint/multinode/train.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def __init__(self, path, project_name, storage_client, model, **kwargs):
+        super().__init__(**kwargs)
+        self.writer = GCSDistributedWriter(path, project_name, storage_client)
+        self.reader = GCSDistributedReader(
+            path, project_name, storage_client)
+        self.checkpoint_io = DatafluxLightningCheckpoint(
+            project_name, storage_client)
+        self.model = model
+        self.storage_client = storage.Client(project=project_name,)
```
jdnurme, line 38: nit: remove the comma after project_name if its the last kwarg

demo/lightning/checkpoint/multinode/train.py
```diff
@@ -0,0 +1,237 @@
...
+                checkpoint.pop("optimizer_states", []))
+        })
+        save(converted_state, checkpoint_id=path, storage_writer=self.writer)
+
+        if self.global_rank == 0:
+            self.checkpoint_io.save_checkpoint(checkpoint,
+                                               path / _METADATA_FILENAME)
+
+    def get_sharded_state_dict_context(self, module: Module) -> Generator[None, None, None]:
+        from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
```
jdnurme, line 65: Is there a specific reason this is imported within the function? since there is a second import of this on line 83 should we just import this in the global scope?

jdnurme (approved):

demo/lightning/checkpoint/multinode/README.md
jdnurme: I mentioned this above, but hyperlinking to the strategy page would probably clarify this.

demo/lightning/checkpoint/multinode/train.py
Yash9060: Ack will do this in a follow up PR. Filed b/366046586

demo/lightning/checkpoint/multinode/README.md
Yash9060: ACK, will do this in a follow up PR.

demo/lightning/checkpoint/multinode/train.py
Yash9060: Done

demo/lightning/checkpoint/multinode/train.py
Yash9060: Hmm that makes sense. This is sort of copied over from FSDP load_checkpoint implementation and hence I didnt refactor it. 

Moved to global import.

demo/lightning/checkpoint/multinode/README.md
Yash9060: Added a note saying custom strategy needs to extend StrategyClass or one of the class which extends strategy.

demo/lightning/checkpoint/multinode/train.py
Yash9060: Changed.

demo/lightning/checkpoint/multinode/README.md
Yash9060: Good idea. Done

demo/lightning/checkpoint/multinode/README.md
```diff
@@ -0,0 +1,40 @@
...
+  * `CKPT_DIR_PATH`: The full path of the directory in which to save checkpoints, in the format `gs://<bucket>/<directory>/`
+2. Set the optional environment variables, if desired:
+  * `NUM_LAYERS`: The number of layers in the model, which affects the size of the model and therefore the size of the checkpoints
+  * `ACCELERATOR`: Set to `gpu` if running on a GPU, or `cpu` if running on a CPU (default)
+    * If running on a GPU, you also must set `PJRT_DEVICE` to `CPU`.
+  * `TRAIN_STRATEGY`: This demo requires custom strategy which should extend DDP Strategy or FSDP Strategy. Extending FSDP Strategy requires GPU.
```
MattIrv, line 18: My previous comment here was really about what values you can set for `TRAIN_STRATEGY`. Is `fsdp` the only valid value or is there something else? If there's just the one, it should be set by default and shouldn't be included here

demo/lightning/checkpoint/multinode/train.py
MattIrv: Don't see this addressed

demo/lightning/checkpoint/multinode/train.py
```diff
@@ -0,0 +1,237 @@
...
+                          project_name=project,
+                          storage_client=None,
+                          model=model,
+                          state_dict_type="sharded",
+                      ),
+                      num_nodes=2
+                      )
+    trainer.fit(model, dataloader)
+    # new_path = ckpt_dir_path + \
+    #     "lightning_logs/version_0/checkpoints/checkpoint-epoch=00-step=03.ckpt"
```
MattIrv, line 196: Remove commented code. Does this work correctly without setting new_path here?

demo/lightning/checkpoint/multinode/README.md
MattIrv, line 1: Can you update the title of this PR to be something more descriptive?

MattIrv (approved):

demo/lightning/checkpoint/multinode/README.md
Yash9060: It can be anything which extends Strategyclass (or any class which extends Strategy like FSDP/ DDP) Hence I think it is better to keep this as a variable.

demo/lightning/checkpoint/multinode/train.py
Yash9060: Removed. Yep the code works fine without this.

demo/lightning/checkpoint/multinode/README.md
Yash9060: Updated. I think I just forgot to change the title and took and the default one.

demo/lightning/checkpoint/multinode/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Running locally
+
+1. Set the environment variables required to run the demo. These include:
+  * `PROJECT`: The GCP project you are using
+  * `CKPT_DIR_PATH`: The full path of the directory in which to save checkpoints, in the format `gs://<bucket>/<directory>/`
+2. Set the optional environment variables, if desired:
+  * `NUM_LAYERS`: The number of layers in the model, which affects the size of the model and therefore the size of the checkpoints
+  * `ACCELERATOR`: Set to `gpu` if running on a GPU, or `cpu` if running on a CPU (default)
+    * If running on a GPU, you also must set `PJRT_DEVICE` to `CPU`.
+  * `TRAIN_STRATEGY`: This demo requires custom strategy which should extend [Strategy class](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.strategies.Strategy.html) or one of the class which extend Strategy class. When using Strategy which extends [FSDPStrategy](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.strategies.FSDPStrategy.html), GPU is required.
```
Yash9060, line 18: @MattIrv can you check if this looks correct ?

demo/lightning/checkpoint/multinode/train.py
Yash9060: I was yet to push the change. Done

demo/lightning/checkpoint/multinode/README.md
Yash9060: Removed this from readme.

PR #107, opened 2024-09-12
Parameterizing min_epochs, max_epochs & max_steps

Yash9060:
Internal issues: b/366046586
One of the comment on https://github.com/GoogleCloudPlatform/dataflux-pytorch/pull/106 was to parameterize min_epochs/max_epochs/max_steps.  Hence this PR achieves that. Also updated documentation and example-deployment.yaml for single node & multi node. 
- [X] Tests pass

### DIFF
--- a/demo/lightning/checkpoint/multinode/README.md
+++ b/demo/lightning/checkpoint/multinode/README.md
@@ -15,6 +15,12 @@ The code in this folder provides a training demo for multi node checkpointing wi
   * `NUM_LAYERS`: The number of layers in the model, which affects the size of the model and therefore the size of the checkpoints
   * `ACCELERATOR`: Set to `gpu` if running on a GPU, or `cpu` if running on a CPU (default)
     * If running on a GPU, you also must set `PJRT_DEVICE` to `CPU`.
+  * `MIN_EPOCHS_SAVE`: Minimum epochs for which training should run during first training loop which saves checkpoint. Defaults to None. For detailed explaination of min_epochs see [here](https://lightning.ai/docs/pytorch/stable/common/trainer.html#min-epochs).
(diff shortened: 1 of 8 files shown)


demo/lightning/checkpoint/multinode/README.md
```diff
@@ -15,6 +15,12 @@ The code in this folder provides a training demo for multi node checkpointing wi
...
+  * `MIN_EPOCHS_RESTORE`: Minimum epochs for which training should run during second training loop which restores checkpoint. Defaults to None.
+  * `MAX_EPOCHS_SAVE` : Maximum epochs for which training should run during first training loop, which saves checkpoint. Defaults to None. For detailed explanation of max_epochs see [here](https://lightning.ai/docs/pytorch/stable/common/trainer.html#max-epochs).
+  * `MAX_EPOCHS_RESTORE` : Maximum epochs for which training should run during second training loop, which restores checkpoint. Defaults to None.
+  * `MAX_STEPS_SAVE`: Maximum number of steps for which training can run during first trainig loop. Defaults to None (no limit). For more infomration on max_steps see [here](https://lightning.ai/docs/pytorch/stable/common/trainer.html#max-steps).
+  * `MAX_STEPS_RESTORE`: Maximum number of steps for which training can run during second trainig loop. Defaults to None (no limit).
```
MattIrv, line 23: For all of these, would it make more sense to default to the previous hardcoded value instead of None or unlimited?

MattIrv (approved):

demo/lightning/checkpoint/multinode/README.md
Yash9060: I have been wondering about that as well. It will be easier to for everyone to run the quick demo instead of setting 10 environment variables. 

The only reason I did not keep these is for semantic reason since those values does not represent the "true default" value. 

Let me change the default values to previous vals.

PR #109, opened 2024-09-13
Add a helper method to create storage_client when needed.

awonak:
Additional fixes for #58, ensure a `storage_client` is present when needed.
Internal issue b/366060916

We want the DataFluxMapStyleDataset constructor paramenter `storage_client` to be optional due to #58, however we need a storage client present when calling `dataflux_core.download.download_single`, etc. Therefore, we should dynamically construct one when needed if one wasn't provided in the constructor.

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/dataflux_mapstyle_dataset.py
@@ -138,11 +138,10 @@ class DataFluxMapStyleDataset(data.Dataset):
         self.bucket_name = bucket_name
         self.data_format_fn = data_format_fn
         self.config = config
-        # If composed download is enabled an a storage_client was provided,
+        # If composed download is enabled and a storage_client was provided,
         # check if the client has permissions to create and delete the
         # composed object.
-        if storage_client is not None \
-                and self.config.max_composite_object_size != 0:
+        if storage_client is not None and self.config.max_composite_object_size:
             missing_perm = _get_missing_permissions(
                 storage_client=self.storage_client,
                 bucket_name=self.bucket_name,
@@ -223,12 +222,11 @@ class DataFluxMapStyleDataset(data.Dataset):
     def _get_or_create_storage_client(self):
         """Provide initialized storage client or construct one.
 
-        Some environments do not support pickling client objects (#58), so
+        Some environments do not support pickling client objects, so
         storage_client is an optional parameter. If the dataset was
         initialized without a storage client, construct and provide one for
         use.
         """
         if self.storage_client is not None:
             return self.storage_client
-        else:
-            return storage.Client(project=self.project_name)
+        return storage.Client(project=self.project_name)
--- a/dataflux_pytorch/dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/dataflux_mapstyle_dataset.py
@@ -20,9 +20,11 @@ import os
 import warnings
 
 import dataflux_core
-from dataflux_pytorch._helper import _get_missing_permissions
+from google.cloud import storage
 from google.cloud.storage.retry import DEFAULT_RETRY
 from torch.utils import data
+
+from dataflux_pytorch._helper import _get_missing_permissions
 
 MODIFIED_RETRY = DEFAULT_RETRY.with_deadline(100000.0).with_delay(
     initial=1.0, multiplier=1.5, maximum=30.0)
@@ -117,8 +119,8 @@ class DataFluxMapStyleDataset(data.Dataset):
             data_format_fn: A function that formats the downloaded bytes to the desired format.
                 If not specified, the default formatting function leaves the data as-is.
             storage_client: The google.cloud.storage.Client object initiated with sufficient permission
-                to access the project and the bucket. If not specified, it will be created
-                during initialization.
+                to access the project and the bucket. If not specified, one will be created
+                when needed.
 
         Returns:
             None.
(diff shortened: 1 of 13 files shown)


dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -136,8 +138,11 @@ def __init__(
         self.bucket_name = bucket_name
         self.data_format_fn = data_format_fn
         self.config = config
-        # If composed download is enabled, check if the client has permissions to create and delete the composed object.
-        if self.config.max_composite_object_size != 0:
+        # If composed download is enabled an a storage_client was provided,
```
jdnurme, line 141: nit: "and a storage_client"

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -136,8 +138,11 @@ def __init__(
...
+        # check if the client has permissions to create and delete the
+        # composed object.
+        if storage_client is not None \
+                and self.config.max_composite_object_size != 0:
```
jdnurme, line 145: to be pythonic, this line can just be `and self.config.max_compose_object_size` as zero will default to false and any non-zero value will default to true

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -214,3 +219,16 @@ def _list_GCS_blobs_with_retry(self):
...
+        """Provide initialized storage client or construct one.
+
+        Some environments do not support pickling client objects (#58), so
+        storage_client is an optional parameter. If the dataset was
+        initialized without a storage client, construct and provide one for
+        use.
+        """
+        if self.storage_client is not None:
+            return self.storage_client
+        else:
```
jdnurme, line 233: This else block can be deleted given the preceding if is a return.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -97,6 +98,67 @@ def test_init_with_required_parameters(self, mock_dataflux_core):
...
+        mock_listing_controller.run.return_value = self.want_objects
+        mock_dataflux_core.fast_list.ListingController.return_value = (
+            mock_listing_controller)
+
+        # Act.
+        ds = dataflux_mapstyle_dataset.DataFluxMapStyleDataset(
+            project_name=self.project_name,
+            bucket_name=self.bucket_name,
+            # Ensure the dataset can be constructed without setting storage_client.
+            # storage_client=self.storage_client,
```
Yash9060, line 115: Remove this line ?

jdnurme (approved):

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -97,6 +98,67 @@ def test_init_with_required_parameters(self, mock_dataflux_core):
...
+        )
+
+        # Assert.
+        self.assertEqual(
+            ds.objects,
+            self.want_objects,
+            f"got listed objects {ds.objects}, want {self.want_objects}",
+        )
+
+        # Ensure the dataset can be constructed without setting storage_client (#58).
```
Yash9060, line 125: Change the comment to not reference line # ? Since the line number might change in future and might be confusing for others to understand.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
jdnurme: I think for the clarity of the test it's ok to leave this line, as it indicates we're testing its omission.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -97,6 +98,67 @@ def test_init_with_required_parameters(self, mock_dataflux_core):
...
+        mock_listing_controller.run.return_value = self.want_objects
+        mock_dataflux_core.fast_list.ListingController.return_value = (
+            mock_listing_controller)
+
+        # Act.
+        ds = dataflux_mapstyle_dataset.DataFluxMapStyleDataset(
+            project_name=self.project_name,
+            bucket_name=self.bucket_name,
+            # Ensure the dataset can be constructed without setting storage_client.
+            # storage_client=self.storage_client,
```
Yash9060, line 145: Same as above ? (I dont think we need the this commented line 

`storage_client = self.storage_client`

Yash9060 (approved):

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
jdnurme: This is referencing the static issue number here: https://github.com/GoogleCloudPlatform/dataflux-pytorch/issues/58, though i think it's ok to just reference that in the PR description and not inline with the code.

Yash9060: I agree. Descriptions seems much better fit for this than the code.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
Yash9060: I would have just preferred the comment without the "code". But I guess it works either way.

PR #110, opened 2024-09-16
Make step time configurable

abhibyreddi:
- [X] Tests pass; tested manually with different sleep times and confirmed that total time increased with higher step_time

### DIFF
--- a/demo/lightning/image-segmentation/model.py
+++ b/demo/lightning/image-segmentation/model.py
@@ -15,6 +15,7 @@
  """
 
 import torch
+import time
 import lightning.pytorch as pl
 from unet3d import Unet3D
 from losses import DiceCELoss
@@ -26,6 +27,7 @@ class Unet3DLightning(pl.LightningModule):
         super().__init__()
         self.flags = flags
         self.benchmark = flags.benchmark
+        self.step_time = flags.step_time
         self.model = Unet3D(1,
                             3,
                             normalization=flags.normalization,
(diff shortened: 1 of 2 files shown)


demo/lightning/image-segmentation/model.py
```diff
@@ -58,6 +60,7 @@ def training_step(self, train_batch, batch_idx):
         if self.benchmark:
             # Returning None will break distributed training, so
             # return a scalar of value 1 that represents loss to skip training.
+            time.sleep(self.step_time)
```
MattIrv, line 63: This approach is probably ok for now and for our benchmarks, but try to measure the actual step time if you're running it this way since it will end up being longer than the configured step time.

The approach for the Tessellations benchmarks is for this to control essentially a "minimum step time" which starts counting from when the step begins (i.e. from the very start of training_step) and then sleeps only on rank 0 if there is still time needed to get to the step time. Sleeping on all ranks can make the actual step time longer since they may not start sleeping at the exact same time. Additionally, I think that the configured step time for those benchmarks may include the data loading time (i.e. the time it takes to get train_batch) but we'd need to double check with Bernard or the team to see if that's really true.

But for getting our quick numbers documented, I think this approach is fine for now.

MattIrv (approved):

demo/lightning/image-segmentation/model.py
abhibyreddi: Noted. Two changes to make as discussed in our meeting today:
1. Sleep only if the node's rank is 0
2. Estimate the actual step time from total training time or per-epoch time. Document that instead of/along with of the configured `step_time`

PR #111, opened 2024-09-16
Remove client initialization for fast listing from dataflux-pytorch

akansha1812:
To fix [#58](https://github.com/GoogleCloudPlatform/dataflux-pytorch/issues/58)

When multiprocess starts with spawn or forkserver,  passing storage client to fast-listing from dataflux pytorch will result in pickling error. Multi-processing fails as it cannot serialize/pickle storage client. 
- remove storage client creation for fast listing from dataflux_pytorch/dataflux_mapstyle_dataset.py.

The change got add in PR [#109]( https://github.com/GoogleCloudPlatform/dataflux-pytorch/pull/109)

- [ ] Tests pass

### DIFF
--- a/dataflux_pytorch/dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/dataflux_mapstyle_dataset.py
@@ -200,7 +200,8 @@ class DataFluxMapStyleDataset(data.Dataset):
                     prefix=self.config.prefix,
                     retry_config=self.config.list_retry_config,
                 )
-                lister.client = self._get_or_create_storage_client()
+                if self.storage_client is not None:
+                    lister.client = self.storage_client
                 listed_objects = lister.run()
 
             except Exception as e:
--- a/dataflux_pytorch/dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/dataflux_mapstyle_dataset.py
@@ -200,6 +200,7 @@ class DataFluxMapStyleDataset(data.Dataset):
                     prefix=self.config.prefix,
                     retry_config=self.config.list_retry_config,
                 )
+                # If the dataset was not initialized with an storage_client, ensure that we do not attach a client to the lister to avoid pickling errors (#58).
                 if self.storage_client is not None:
                     lister.client = self.storage_client
                 listed_objects = lister.run()
--- a/dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
@@ -21,7 +21,7 @@ from unittest import mock
 
 from google.cloud import storage
 
-from dataflux_core.tests import fake_gcs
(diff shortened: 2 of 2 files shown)


dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -200,7 +200,8 @@ def _list_GCS_blobs_with_retry(self):
                     prefix=self.config.prefix,
                     retry_config=self.config.list_retry_config,
                 )
-                lister.client = self._get_or_create_storage_client()
+                if self.storage_client is not None:
+                    lister.client = self.storage_client
```
MattIrv, line 204: Thanks for noticing and fixing this! A few questions:
1. Are you sure this is the only place in the code that is affected? Will there be a similar error for the download parts of the code?
2. I thought when this was fixed before we had added tests that would have caught this. Is there any additional test we can add that would help prevent this from being reintroduced?

MattIrv (approved):

dataflux_pytorch/dataflux_mapstyle_dataset.py
akansha1812: Adding storage_client to DataFluxMapStyleDataset object was causing pickling error as it was used by pytorch dataloader. pytorch dataloader calls multiprocessing.  With _get_or_create_storage_client method, the client is not added to DataFluxMapStyleDataset object. 

In dataflux_mapstyle_dataset.py we are calling download_single and threaded_download which does not call multiprocessing.  Multiprocessing is called only by parallel_composed_download.

Also ran a test and looks like this was the only place it will fail. 

Our unit test wasn't covering listing,  I'll add more unit tests to cover this.

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -200,7 +200,8 @@ def _list_GCS_blobs_with_retry(self):
                     prefix=self.config.prefix,
                     retry_config=self.config.list_retry_config,
                 )
-                lister.client = self._get_or_create_storage_client()
+                if self.storage_client is not None:
```
awonak, line 203:
```suggestion
                # If the dataset was not initialized with an storage_client, ensure that we do not attach a client to the lister to avoid pickling errors (#58).
                if self.storage_client is not None:
```

dataflux_pytorch/dataflux_mapstyle_dataset.py
awonak: Thanks for catching this! I misunderstood where the pickling error was being thrown. This looks like a good solution. I added a suggestion for additional comment to make the issue clear.

awonak (approved):

dataflux_pytorch/dataflux_mapstyle_dataset.py
akansha1812: Done.

PR #114, opened 2024-09-18
Multipart checkpoint upload

jdnurme:
Add base code for multipart upload. Followup CL will introduce comprehensive testing at which point we can enable it by default. Multiprocess code written by @andrewsg.

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/multipart_upload/multipart.py
+++ b/dataflux_pytorch/multipart_upload/multipart.py
@@ -72,10 +72,6 @@ def upload_chunks_concurrently_from_bytesio(
     rules, or refer to the XML API documentation linked above to learn more
     about how to list and delete individual downloads.
 
-    Using this feature with multiple threads is unlikely to improve upload
-    performance under normal circumstances due to Python interpreter threading
-    behavior. The default is therefore to use processes instead of threads.
-
     ACL information cannot be sent with this function and should be set
     separately with :class:`ObjectACL` methods.
 
--- a/dataflux_pytorch/multipart_upload/multipart.py
+++ b/dataflux_pytorch/multipart_upload/multipart.py
@@ -273,7 +273,7 @@ class _BufferViewXMLMPUPart(XMLMPUPart):
 
 
 def _buffer_view_upload_part(
-    maybe_pickled_client,
+    client,
     url,
     upload_id,
     view,
@@ -284,16 +284,7 @@ def _buffer_view_upload_part(
     headers,
     retry,
 ):
-    """Helper function that runs inside a thread or subprocess to upload a part.
-
-    `maybe_pickled_client` is either a Client (for threads) or a specially
-    pickled Client (for processes) because the default pickling mangles Client
-    objects."""
-
-    if isinstance(maybe_pickled_client, Client):
-        client = maybe_pickled_client
-    else:
-        client = pickle.loads(maybe_pickled_client)
+    """Helper function that runs inside a thread to upload a part."""
     part = _BufferViewXMLMPUPart(
         url,
         upload_id,
--- a/dataflux_pytorch/multipart_upload/multipart.py
+++ b/dataflux_pytorch/multipart_upload/multipart.py
@@ -1,3 +1,18 @@
+"""
+ Copyright 2024 Google LLC
+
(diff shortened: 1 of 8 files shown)


dataflux_pytorch/multipart_upload/multipart.py
```diff
@@ -0,0 +1,313 @@
...
+    If the upload fails in a way that precludes cancellation, such as a
+    hardware failure, process termination, or power outage, then the incomplete
+    upload may persist indefinitely. To mitigate this, set the
+    `AbortIncompleteMultipartUpload` with a nonzero `Age` in bucket lifecycle
+    rules, or refer to the XML API documentation linked above to learn more
+    about how to list and delete individual downloads.
+
+    Using this feature with multiple threads is unlikely to improve upload
+    performance under normal circumstances due to Python interpreter threading
+    behavior. The default is therefore to use processes instead of threads.
```
awonak, line 77: I wanted to get clarification here, the docs here suggest the default is to use processes, however we're using threads which makes sense for a network bound task. What is the default referring to in this paragraph?

dataflux_pytorch/multipart_upload/multipart.py
```diff
@@ -0,0 +1,309 @@
...
+            self._checksum_object.update(payload)
+
+        part_query = MPU_PART_QUERY_TEMPLATE.format(part=self._part_number,
+                                                    upload_id=self._upload_id)
+        upload_url = self.upload_url + part_query
+        return "PUT", upload_url, payload, self._headers
+
+
+def _buffer_view_upload_part(
+    maybe_pickled_client,
```
andrewsg, line 276: I forgot to do this earlier, but we can change this to "client", delete the "maybe_picked_client" documentation in the docstring, and cut the entire line 293-296 code block that sets client. There's no need for it because maybe_pickled_client is for multiprocessing support only.

Yash9060 (approved):

dataflux_pytorch/multipart_upload/multipart.py
```diff
@@ -0,0 +1,300 @@
...
+    transport = blob._get_transport(client)
+
+    hostname = _get_host_name(client._connection)
+    url = "{hostname}/{bucket}/{blob}".format(hostname=hostname,
+                                              bucket=bucket.name,
+                                              blob=_quote(blob.name))
+
+    base_headers, object_metadata, content_type = blob._get_upload_arguments(
+        client, content_type, filename=None, command="tm.upload_sharded")
+    headers = {**base_headers, **_headers_from_metadata(object_metadata)}
```
MattIrv, line 153: Do we need to make sure the "dataflux" user-agent header is set here, or is it already included in the base_headers?

dataflux_pytorch/multipart_upload/multipart.py
MattIrv, line 1: Are tests for this file coming in a separate CL?

demo/lightning/checkpoint/singlenode/continuous.py
```diff
@@ -39,7 +39,7 @@ def parse_args():
     parser.add_argument("--max-epochs", type=int, default=5)
     parser.add_argument("--max-steps", type=int, default=3)
     parser.add_argument("--num-workers", type=int, default=1)
-    parser.add_argument("--num-layers", type=int, default=2)
+    parser.add_argument("--num-layers", type=int, default=10)
```
MattIrv, line 42: Did you mean to change this?

demo/lightning/checkpoint/singlenode/continuous.py
```diff
@@ -69,7 +69,11 @@ def main():
                       accelerator="cpu",
                       strategy="ddp",
                       num_nodes=1)
+    t1 = time.time()
     trainer.fit(model, dataloader)
+    t2 = time.time()
+    result_time = t2 - t1
+    print(f"Total elapsed time was {result_time} seconds...")
```
MattIrv, line 76: nit: Because the call to trainer.fit includes all the time it takes to initialize and do a step on the model, if you want real timing info it's better to use the checkpoint benchmark code instead of the demo code

MattIrv (approved):

PR #115, opened 2024-09-18
Run unit tests on x86 Mac

abhibyreddi:
This makes Kokoro trigger existing unit tests on a x86 Mac as part of presubmits. 

This change can be checked in after verifying that Kokoro is in fact running the unit tests on the newly configured machine. The corresponding google3 CL (cl/676086380) needs to be checked in prior to that.


MattIrv (approved):

kokoro/build_mac_x86.sh
```diff
@@ -0,0 +1,56 @@
...
+    echo Installing requirements.
+    pip install -r requirements.txt
+
+    echo Installing required dependencies.
+    pip install .
+}
+
+function run_unit_tests() {
+    echo "Running unit tests on MacOS (x86)."
+    python3 -m pytest dataflux_pytorch/tests -vv --junit-xml="${KOKORO_ARTIFACTS_DIR}/unit_tests/sponge_log.xml" --log-cli-level=DEBUG
```
jdnurme, line 50: Is this command different than the existing shell scripts? If it's not unique for the mac-os we can just point the .cfg file to an existing script.

jdnurme (approved):

kokoro/build_mac_x86.sh
akansha1812: +1

MattIrv:
@abhibyreddi this PR has been open for 2 weeks, is it ready to merge or should it be closed?

abhibyreddi:
> @abhibyreddi this PR has been open for 2 weeks, is it ready to merge or should it be closed?

I will check this in today.

kokoro/build_mac_x86.sh
abhibyreddi: The only difference between this and the linux build script is the apt-get commands.

akansha1812 (commented):
test_init_with_spawn_multiproces test  contains real storage client and can be removed as we have added test_list_GCS_blobs_with_spawn_multiprocess.  

The reason for adding real client  to unit is to show that _list_GCS_blobs will fail when multiprocess start method is spawn/forkserver and a client is passed by the user.  

Tests test_list_GCS_blobs_with_spawn_multiprocess makes sure if the client is created  when  multiprocess start method is spawn, then _list_GCS_blobs will fail.

dataflux_pytorch/tests/test_dataflux_iterable_dataset.py
MattIrv, line 1: As discussed in chat, let's keep both of these tests since we have a good workaround for the credentials issue now

MattIrv (approved):

dataflux_pytorch/tests/test_dataflux_iterable_dataset.py
abhibyreddi: Done! Added both of them back. Will submit after confirming that those two are passing.

PR #117, opened 2024-09-19
adds unit tests, adds presubmit integration test, updated demo code

jdnurme:
This PR adds basic unit testing and a small scale presubmit integration test for multipart upload. Updates to the demo were made to use command line variables, and README was updated to match new execution methods.

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/benchmark/README.md
+++ b/dataflux_pytorch/benchmark/README.md
@@ -18,42 +18,28 @@ First ensure you are running within a virtual python enviroment, make sure gclou
 gcloud config set project {PROJECT_ID}
 ```
 
-Then set the enviroment variables.
+Then set the command line variables.
 
-`CKPT_DIR_PATH` is the location of where to save the checkpoints. `STEPS` is the number of steps the model will take (the number of checkpoints created will be the same). The default value for `STEPS` is 5. The benchmark will run `save_checkpoint` repeatedly and produce the average at the end, then run `load_checkpoint` on all the saved checkpoints and produce the average.
+`--ckpt_dir_path` is the location of where to save the checkpoints. `--steps` is the number of steps the model will take (the number of checkpoints created will be the same). The default value for `--steps` is 5. The benchmark will run `save_checkpoint` repeatedly and produce the average at the end, then run `load_checkpoint` on all the saved checkpoints and produce the average.
 
-```shell
-export CKPT_DIR_PATH=`gs://path/to/directory/`
-export STEPS=5
-```
-
-You can also optionally change the size of the model. The `LAYERS` variable will be passed into `nn.Transformer` for `num_encoder_layers` and `num_decoder_layers`. The default value for `LAYERS` is 100.
-
-```shell
-export LAYERS=1000
-```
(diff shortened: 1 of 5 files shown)


dataflux_pytorch/benchmark/README.md
```diff
@@ -18,42 +18,28 @@ First ensure you are running within a virtual python enviroment, make sure gclou
 gcloud config set project {PROJECT_ID}
 ```
 
-Then set the enviroment variables.
+Then set the command line variables.
 
-`CKPT_DIR_PATH` is the location of where to save the checkpoints. `STEPS` is the number of steps the model will take (the number of checkpoints created will be the same). The default value for `STEPS` is 5. The benchmark will run `save_checkpoint` repeatedly and produce the average at the end, then run `load_checkpoint` on all the saved checkpoints and produce the average.
+`--ckpt_dir_path` is the location of where to save the checkpoints. `--steps` is the number of steps the model will take (the number of checkpoints created will be the same). The default value for `--steps` is 5. The benchmark will run `save_checkpoint` repeatedly and produce the average at the end, then run `load_checkpoint` on all the saved checkpoints and produce the average.
```
MattIrv, line 23: nit: it might be helpful to separate each flag on to its own line to make it easier to read and notice each flag that needs to be set

dataflux_pytorch/benchmark/README.md
```diff
@@ -18,42 +18,28 @@ First ensure you are running within a virtual python enviroment, make sure gclou
...
 
 ### Dataflux Lightning Checkpoint
 
-If you are benchmarking Dataflux Lightning Checkpoint, save information regarding your project and make sure to enable the flag by setting it to `1`.
-
-```shell
-export PROJECT=`YOUR_PROJECT_NAME`
-export DATAFLUX_CKPT=1
-```
+If you are benchmarking Dataflux Lightning Checkpoint, save information regarding your project and make sure to enable the `--dataflux_ckpt` flag.
```
MattIrv, line 29: Should this be the default?

kokoro/presubmit.sh
```diff
@@ -57,8 +57,10 @@ function run_unit_tests() {
 }
 
 function run_integration_tests(){
-    echo Running integration tests.
+    echo Running basic integration test.
     python3 -m pytest dataflux_pytorch/integration_tests/integration_test.py -vv --junit-xml="${KOKORO_ARTIFACTS_DIR}/integration_tests/sponge_log.xml" --log-cli-level=DEBUG
+    echo Running checkpoint integration test.
+    python3 dataflux_pytorch/benchmark/lightning_checkpoint_benchmark.py --enable-multipart --project=dataflux-project --ckpt-dir-path=gs://jd-upload-west/ --dataflux-ckpt --layers=10 --steps=5
```
MattIrv, line 63: We may want to use a dedicated bucket for this, with a retention rule enabled that automatically deletes files after a while, so that old checkpoint files don't pile up and take up space

MattIrv (approved):

dataflux_pytorch/benchmark/README.md
jdnurme: yeah i think so, i'll reverse it

kokoro/presubmit.sh
jdnurme: my bucket does have those rules but i'll make a more officially named one for this

PR #122, opened 2024-09-24
Add code to clear kernel cache after saving checkpoints

abhibyreddi:
- [X] Tests pass - ran the script on a VM with and without the new flag and verified it's working as intended

### DIFF
--- a/dataflux_pytorch/benchmark/README.md
+++ b/dataflux_pytorch/benchmark/README.md
@@ -25,6 +25,8 @@ Then set the command line variables.
 
 `--layers`: you can also optionally change the size of the model. The `--layers` argument will be passed into `nn.Transformer` for `num_encoder_layers` and `num_decoder_layers`. The default value for `--layers` is 100.
 
+`--clear-kernel-cache`: this is an optional argument which when set clears kernel cache after saving the checkpoints. If benchmarking gcsfuse, this argument must be set.
+
 ### Dataflux Lightning Checkpoint
 
 `--no-dataflux-ckpt`: If you are not benchmarking Dataflux Lightning Checkpoint, this will disable dataflux checkpointing and use the default lightning checkpointing instead. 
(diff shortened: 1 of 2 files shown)


dataflux_pytorch/benchmark/README.md
```diff
@@ -25,6 +25,8 @@ Then set the command line variables.
 
 `--layers`: you can also optionally change the size of the model. The `--layers` argument will be passed into `nn.Transformer` for `num_encoder_layers` and `num_decoder_layers`. The default value for `--layers` is 100.
 
+`--clear-kernel-cache`: this is an optional argument which when set clears kernel cache after saving the checkpoints. If benchmarking gcsfuse, this argument must be set.
```
MattIrv, line 28: I'd recommend making the "If benchmarking" language a bit more generic. While we ran into this when trying out gcsfuse, it would likely apply to any filesystem, for example when saving to local disk. In fact now that I think about it that's probably why my Local SSD tests I ran before had incredibly fast read times

MattIrv (approved):

dataflux_pytorch/benchmark/README.md
abhibyreddi: Changed it to "If benchmarking local filesystems or storage solutions that expose a filesystem interface (such as gcsfuse), this argument must be set in order to get more accurate performance measurements."

Does that look good?

MattIrv: Looks good. Something else I realized is that will only work on Linux/Mac as-is, and may only work when running with sudo. Might need to document that also.

abhibyreddi: Done! Made it so that the command to clear cache is run only if the OS is Linux or MacOS. Also added a note in the README about that. Does it look good?

MattIrv: Great, thanks

PR #123, opened 2024-09-24
update continuous to run full benchmark

jdnurme:
Updating continuous benchmark to run 1000 layer checkpoint benchmark. Also removing old continuous benchmark as it is no longer needed.

- [x] Tests pass


kokoro/bench.sh
```diff
@@ -68,6 +68,7 @@ function run_benchmarks(){
     python3 -u ./demo/lightning/image-segmentation/train.py --local --benchmark --gcp_project=dataflux-project --gcs_bucket=dataflux-demo-public --images_prefix=image-segmentation-dataset/images --labels_prefix=image-segmentation-dataset/labels --num_nodes=1 --num_devices=5 --epochs=2;
     echo Running single node checkpointing benchmark.
     python3 -u ./demo/lightning/checkpoint/singlenode/continuous.py --project=dataflux-project --ckpt-path=gs://df-ckpt-continuous;
```
awonak, line 70: It looks like this file was deleted, so this line should be deleted too.

Yash9060 (approved):

kokoro/bench.sh
jdnurme: oh huh, i thought i did delete that, good catch that would have exploded.

awonak (approved):

PR #128, opened 2024-09-26
Changing hardcoded values to placeholders

Yash9060:
Changing the hardcoded values to placeholder in benchmark-deploy.yaml

- [X] Tests pass

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/benchmark-deploy.yaml
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/benchmark-deploy.yaml
@@ -15,7 +15,7 @@
 apiVersion: jobset.x-k8s.io/v1alpha2
 kind: JobSet
 metadata:
-  name: yashsha-checkpoint
+  name: my-job-run
   labels:
     kueue.x-k8s.io/queue-name: multislice-queue  # Name of the LocalQueue
   annotations:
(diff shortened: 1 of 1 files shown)


dataflux_pytorch/benchmark/checkpointing/multinode/benchmark-deploy.yaml
```diff
@@ -15,7 +15,7 @@
 apiVersion: jobset.x-k8s.io/v1alpha2
 kind: JobSet
 metadata:
-  name: yashsha-checkpoint
+  name: my-job-run
```
awonak, line 18: I see in other areas it looks like you can use variables such as `$(JOBSET_NAME)`. Would it make sense to use variables instead of placeholders? That would help make it more explicit what values the user need to provide instead of interpreting these values as placeholders.

Yash9060: $(JOBSET_NAME) is something which will be replaced by GKE automatically, whereas these are the values which customer needs to change explicitly, but we can do something like `<my_job_run>` since `$()` is reserved for automated replacements.

Yash9060: Done

data/wikitext-2.txt
MattIrv, line 1: Did you mean to check this in?

MattIrv (approved):

awonak (approved):

PR #129, opened 2024-09-26
Print average times to save and load checkpoints together

abhibyreddi:
- [X] Tests pass - ran on a multi-node GKE cluster and verified that the output looks as expected
- [X] Appropriate changes to documentation are included in the PR - no changes to documentation needed

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
@@ -7,13 +7,15 @@ import statistics
 
 from lightning import Trainer
 from lightning.pytorch.callbacks import ModelCheckpoint
-from lightning.pytorch.demos import (
-    WikiText2)
+from lightning.pytorch.demos import (WikiText2)
 import torch.distributed
 from torch.utils.data import DataLoader
 
 
-def main(project: str, ckpt_dir_path: str, save_only_latest: bool, ckpt_restore_path: str = ""):
+def main(project: str,
+         ckpt_dir_path: str,
+         save_only_latest: bool,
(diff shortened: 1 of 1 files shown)


abhibyreddi:
This makes it easy for me to grab the save and load times from the logs.

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -42,32 +44,29 @@ def main(project: str, ckpt_dir_path: str, save_only_latest: bool, ckpt_restore_
...
+    )
     trainer.fit(model, dataloader)
     start = time.time()
     for i in range(max_steps_save):
-        trainer.save_checkpoint(os.path.join(
-            ckpt_dir_path, f'checkpoints/ckpt_{i}.ckpt/'))
+        trainer.save_checkpoint(
+            os.path.join(ckpt_dir_path, f'checkpoints/ckpt_{i}.ckpt/'))
     end = time.time()
-    if torch.distributed.get_rank() == 0:
```
MattIrv, line 62: It may still be worth printing something here (or even after each individual checkpoint is saved/loaded) so that it's clear to the user whether the benchmark is making progress when they run it

MattIrv (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: Agreed! I will put print statements in appropriate places to log progress. I've made a couple more minor but necessary changes to this file which I want to include in a different PR. I will add the print statements too to the new set of commits.

akansha1812 (approved):

PR #131, opened 2024-09-27
update lightning bench with multipart and 10k info

jdnurme:
Update README with findings from full benchmark pass with multipart, plus 10k passes with default, non-multipart, and multipart upload/download settings. Note that download does not include a multipart entry since the performance improvements only apply to upload as of today.

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/singlenode/README.md
+++ b/dataflux_pytorch/benchmark/checkpointing/singlenode/README.md
@@ -103,19 +103,7 @@ Dataflux's implementation of CheckpointIO for PyTorch Lightning is undergoing ac
    </td>
   </tr>
   <tr>
-   <td style="background-color: #f3f3f3">Dataflux (non-multipart)
-   </td>
-   <td style="background-color: #f3f3f3">10
-   </td>
-   <td style="background-color: #f3f3f3">75.6
-   </td>
-   <td style="background-color: #f3f3f3">0.74
-   </td>
-   <td style="background-color: #f3f3f3">102.16
-   </td>
-  </tr>
-  <tr>
-   <td style="background-color: #f3f3f3">Dataflux (multipart)
+   <td style="background-color: #f3f3f3">Dataflux
    </td>
    <td style="background-color: #f3f3f3">10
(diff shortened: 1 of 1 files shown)


abhibyreddi (approved):

dataflux_pytorch/benchmark/checkpointing/singlenode/README.md
```diff
@@ -106,7 +103,7 @@ Dataflux's implementation of CheckpointIO for PyTorch Lightning is undergoing ac
    </td>
   </tr>
   <tr>
-   <td style="background-color: #f3f3f3">Dataflux
+   <td style="background-color: #f3f3f3">Dataflux (non-multipart)
```
MattIrv, line 106: Since multipart is the default and we don't have any scenarios where we recommend disabling multipart (right?) let's just present the multipart numbers as "Dataflux" and remove the non-multipart numbers. What do you think? We don't want to overwhelm users with information and make it harder to understand what performance to expect out of the box

MattIrv (approved):

dataflux_pytorch/benchmark/checkpointing/singlenode/README.md
jdnurme: sounds good to me, will do

dataflux_pytorch/benchmark/checkpointing/singlenode/README.md
```diff
@@ -158,11 +154,35 @@ Dataflux's implementation of CheckpointIO for PyTorch Lightning is undergoing ac
    </td>
    <td style="background-color: #f3f3f3">1000
    </td>
-   <td style="background-color: #f3f3f3">2500
+   <td style="background-color: #f3f3f3">2,634
```
MattIrv, line 157: Why are the file sizes different now? Let's also be consistent with whether or not we're using commas, here you're using one but in the 1000 number above we're not

jdnurme: Updated to include commas everywhere

PR #137, opened 2024-10-02
add option to run benchmark without lightning

jdnurme:
For testing benchmark data without lightning, you can now use the --no-lightning flag. Performance numbers do not significantly differ from lightning implementation (as expected).

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/singlenode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/singlenode/train.py
@@ -28,7 +28,8 @@ from lightning.pytorch.plugins.io import TorchCheckpointIO
 from torch import Tensor
 from torch.utils.data import DataLoader
 
-from dataflux_pytorch.lightning import DatafluxLightningCheckpoint
+from dataflux_pytorch.lightning import DatafluxLightningCheckpoint, path_utils
+from dataflux_pytorch import dataflux_checkpoint
 
 
 class LightningTransformer(LightningModule):
@@ -76,6 +77,7 @@ def parse_args():
     parser.add_argument("--clear-kernel-cache",
                         action="store_true",
                         default=False)
+    parser.add_argument("--no-lightning", action="store_true", default=False)
     return parser.parse_args()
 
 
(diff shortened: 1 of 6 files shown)


dataflux_pytorch/benchmark/checkpointing/singlenode/train.py
```diff
@@ -132,26 +134,52 @@ def main():
...
-            os.path.join(args.ckpt_dir_path, f'ckpt_{i}.ckpt'))
-    end = time.time()
-    print("Average time to load one checkpoint: " +
-          str((end - start) / args.steps) + " seconds")
+    if args.no_lightning:
+        # Create non-lightning checkpoint.
+        bucket_name, ckpt_path = path_utils.parse_gcs_path(args.ckpt_dir_path)
+        ckpt = dataflux_checkpoint.DatafluxCheckpoint(
+            project_name=args.project, bucket_name=bucket_name)
+        CKPT_PATH = ckpt_path if ckpt_path else "checkpoints/epoch0.ckpt"
```
MattIrv, line 142: Should we add the step number to this file name so that it's not overwriting the same file repeatedly? There are GCS limits for how quickly you can rewrite the same object

dataflux_pytorch/benchmark/checkpointing/singlenode/train.py
```diff
@@ -132,26 +134,52 @@ def main():
...
-    print("Average time to save one checkpoint: " +
-          str((end - start) / args.steps) + " seconds")
-    start = time.time()
-    for i in range(args.steps):
-        data = ckpt.load_checkpoint(
-            os.path.join(args.ckpt_dir_path, f'ckpt_{i}.ckpt'))
-    end = time.time()
-    print("Average time to load one checkpoint: " +
-          str((end - start) / args.steps) + " seconds")
+    if args.no_lightning:
```
MattIrv, line 137: Would it be possible to reduce the code duplication at all? One way might be to put if statements only inside each of the `for i in range(args.steps):` loops that determine how exactly the checkpoint is saved/loaded

MattIrv (approved):

dataflux_pytorch/benchmark/checkpointing/singlenode/train.py
jdnurme: Good idea, will add.

Yash9060 (approved):
Update readme.md ?

dataflux_pytorch/benchmark/checkpointing/singlenode/train.py
```diff
@@ -103,6 +105,9 @@ def main():
     args = parse_args()
     if args.steps < 1:
         raise ValueError("Steps need to greater than 0.")
+    if args.no_lightning and args.no_dataflux_ckpt:
+        raise ValueError(
+            "Cannot specify --no-dataflux-ckpt and --no-lightning, pick one.")
```
awonak, line 110: We will have a slight merge conflict with my args change in PR #135. I don't think it'll make sense to add `no-lightning` to the new checkpoint choice flag, so this block will likely be changed to something like:

```python
if args.checkpoint == "no-dataflux" and args.no_lightning:
    ...
```
    

https://github.com/GoogleCloudPlatform/dataflux-pytorch/pull/135/files#diff-daaa477a49e483154fd42d8638792f354f98cd26b79f5c5c4542582757b56fbbR85-R88

jdnurme: yeah i saw this on your PR, will update

awonak (approved):
LGTM. Nice work providing constants for the arg choices!

PR #143, opened 2024-10-07
CPU simulated benchmarking for GKE cluster.

Yash9060:
Simulated CPU benchmarking code for GKE cluster. I have tested this on GKE cluster of 20 N2-standard-64. 
Example of deployment-yaml: http://gpaste/6024689370791936

Using linear model here since it is easier to divide it amongst the processes (in order to ensure consistent size across the processes) . 

Also not using trainer object (& trainer.fit) since training does consume some time (decreases dev velocity) and also produces checkpoints of varying sizes. 

I will have a separate version of similar code which can be run on cloudtop for quick testing. 

- [X] Tests pass


demo/lightning/checkpoint/multinode/train2.py
MattIrv, line 1: Did you mean to remove this?

MattIrv (approved):

jdnurme (approved):

awonak (approved):
LGTM

PR #144, opened 2024-10-07
Add support for multi-node checkpointing with fsspec

abhibyreddi:
- [X] Tests pass - deployed workload on a 2 node GKE cluster that finished successfully
- [ ] Appropriate changes to documentation are included in the PR - follow up PR will include updates to the README file (with instructions on how to run this code on a multi-node cluster)


abhibyreddi:
changes to `demo/lightning/checkpoint/multinode/train.py` are formatting related only

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
```diff
@@ -0,0 +1,109 @@
+import os
```
Yash9060, line 1: Can we move this entire file to demo and an example in demo and use that demo as base in benchmarks train.py 

I think this would make it consistent with multi node benchmarks.

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
```diff
@@ -0,0 +1,109 @@
...
+                        storage_options=None) -> None:
+        if storage_options is not None:
+            raise TypeError(
+                "`FSDPStrategy.save_checkpoint(..., storage_options=...)` is not supported because"
+                " `FSDPStrategy` does not use the `CheckpointIO`.")
+
+        writer = FF.FsspecWriter(self.path, sync_files=False)
+        _ = Path(self.broadcast(filepath))
+
+        # self.broadcast(filepath)
```
Yash9060, line 37: Remove

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
```diff
@@ -0,0 +1,109 @@
...
+    def save_checkpoint(self,
+                        checkpoint,
+                        filepath,
+                        storage_options=None) -> None:
+        if storage_options is not None:
+            raise TypeError(
+                "`FSDPStrategy.save_checkpoint(..., storage_options=...)` is not supported because"
+                " `FSDPStrategy` does not use the `CheckpointIO`.")
+
+        writer = FF.FsspecWriter(self.path, sync_files=False)
```
Yash9060, line 34: Any specific reason for initializing writer in the save_checkpoint instead of init ?

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -4,18 +4,33 @@
...
 from torch.utils.data import DataLoader
 
+DF_FSDP_STRATEGY = "dataflux_fsdp"
+FSDP_STRATEGY = "fsdp"
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--strategy',
+                        choices=[DF_FSDP_STRATEGY, FSDP_STRATEGY],
```
Yash9060, line 24: Did you try passing in non default option in GKE yaml file ? 

I thought we needed some work to support parser.args in GKE and hence demo has flags and GKE version has environment variables.

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
abhibyreddi: Not really. I had to initialize the reader in `load_checkpoint` because the checkpoint path is different for different for different reads. I moved the initialization of the writer to `save_checkpoint` just so it's consistent.

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
abhibyreddi: Done!

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: I have not! But the image segmentation workload had no trouble parsing args from flags. It should work here too but I will double check.

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
abhibyreddi: Done.

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: Made the dataflux fsdp strategy to be the default option. Re-ran the workload by passing `--strategy=fsspec_fsdp` and made sure it's running with no errors.

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
awonak: I agree that it would be better to initialize your reader & writer as instance variables inside the init method so you can reuse them instead of recreating them each time save/load is called.

demo/lightning/checkpoint/multinode/fsspecfsdp.py
```diff
@@ -0,0 +1,107 @@
...
+                        checkpoint,
+                        filepath,
+                        storage_options=None) -> None:
+        if storage_options is not None:
+            raise TypeError(
+                "`FSDPStrategy.save_checkpoint(..., storage_options=...)` is not supported because"
+                " `FSDPStrategy` does not use the `CheckpointIO`.")
+
+        writer = FF.FsspecWriter(self.path, sync_files=False)
+        _ = Path(self.broadcast(filepath))
```
jdnurme, line 35: What is this doing? Is there a reason we're not using the path but still trying to get the file path?

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -81,13 +104,21 @@ def main(project: str,
...
+        if args.strategy == DF_FSDP_STRATEGY:
+            print("Using DatafluxFSDPStrategy")
+            strategy = DatafluxFSDPStrategy(
+                path=new_path,
+                project_name=project,
+                storage_client=None,
+                model=model,
+                state_dict_type="sharded",
+            )
+        else:
```
jdnurme, line 117: I would probably still evaluate the strategy here, and have a final else block raising an error if the strategy is invalid.

demo/lightning/checkpoint/multinode/fsspecfsdp.py
```diff
@@ -0,0 +1,107 @@
...
+            module=module,
+            state_dict_type=StateDictType.SHARDED_STATE_DICT,
+            state_dict_config=state_dict_config,
+            optim_state_dict_config=optim_state_dict_config,
+        )
+        return state_dict_type_context  # type: ignore[return-value]
+
+    def load_checkpoint(self, checkpoint_path):
+        # broadcast the path from rank 0 to ensure all the states are loaded from a common path
+        _ = Path(self.broadcast(checkpoint_path))
```
jdnurme, line 68: Same question about the empty path, this looks like its used directly [in the normal fsdp implementation](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/strategies/fsdp.py#L591).

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
abhibyreddi: I can initialize the writer in init but am lost on how to do the same for reader. The reader needs to be passed the path to the checkpoint that we want to load which is an argument to `FSSpecFSDPStrategy.load_checkpoint` so it's easier to initialize it there. Any tips?

demo/lightning/checkpoint/multinode/fsspecfsdp.py
```diff
@@ -0,0 +1,107 @@
...
+                    )
+                    optim.load_state_dict(flattened_osd)
+
+        # Load metadata (anything not a module or optimizer)
+        new_path = os.path.join(checkpoint_path, _METADATA_FILENAME)
+        metadata = None
+        with reader.fs.create_stream(path=new_path,
+                                     mode='rb') as metadata_file:
+            metadata = torch.load(metadata_file)
+        return metadata
```
jdnurme, line 107: Why does this function return just metadata? Looking at the pytorch FSDP implementation the load function seems to return the [full checkpoint](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/strategies/fsdp.py#L674).

demo/lightning/checkpoint/multinode/fsspecfsdp.py
```diff
@@ -0,0 +1,107 @@
...
+from lightning.pytorch.strategies import FSDPStrategy
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+
+from lightning.pytorch.strategies.fsdp import _METADATA_FILENAME
+from torch.distributed.checkpoint import save, load
+from torch.nn import Module
+from torch.distributed.checkpoint import _fsspec_filesystem as FF
+
+
+class FSSpecFSDPStrategy(FSDPStrategy):
```
awonak, line 18: This class appears to have a lot of duplication with the `DatafluxFSDPStrategy`. I'd suggest thinking about what is different, and how you can abstract that away and have your FSSpecFSDPStrategy inherit DatafluxFSDPStrategy. Here's an (untested) example derived from some similar work I'm doing:

```python

class DatafluxFSDPStrategy(FSDPStrategy):

    def _init_io(self):
        self.writer = GCSDistributedWriter(path, project_name, storage_client)
        self.reader = GCSDistributedReader(path, project_name, storage_client)
        self.checkpoint_io = DatafluxLightningCheckpoint(
            project_name, storage_client)

    def _save(self, converted_state, path):
        save(converted_state, checkpoint_id=path, storage_writer=self.writer)

    def _write_checkpoint(self, checkpoint, path):
        self.checkpoint_io.save_checkpoint(checkpoint, path / _METADATA_FILENAME)

    def save_checkpoint(self, checkpoint, filepath, storage_options=None) -> None:
        if storage_options is not None:
            raise TypeError(...)

        path = Path(self.broadcast(filepath))

        converted_state = {"model": checkpoint.pop("state_dict")}
        converted_state.update({
            f"optimizer_{idx}": optim_state
            for idx, optim_state in enumerate(
                checkpoint.pop("optimizer_states", []))
        })
        self._save(converted_state, path)

        if self.global_rank == 0:
            self._write_checkpoint()


class FSSpecFSDPStrategy(DatafluxFSDPStrategy):

    def _init_io(self):
        self.writer = FF.FsspecWriter(self.path, sync_files=False)
        self.reader = FF.FsspecWriter(self.path, sync_files=False)

    def _write_checkpoint(self, path):
        bucket = gcsfs.GCSFileSystem()
        with bucket.open(os.path.join(filepath, _METADATA_FILENAME), 'wb') as f:
            torch.save(checkpoint, f)
```

This shows how we isolate the areas that change and move them into a private method that we can override with the changes specific to our different strategy. This also helps make the demo more readable as it highlights what needs to change from the default implementation.

demo/lightning/checkpoint/multinode/fsspecfsdp.py
abhibyreddi: Path() converts the path passed to it to a posix path. For example, Path("gs://checkpoint-benchmark-west1") would return "gs:/checkpoint-benchmark-west1" which would break things when fsspec tries to read/write to/from the bucket (but OK for the normal FSDP because it assumes local paths). But I think we would still need to broadcast the patch to all nodes. 

It's probably okay to just do `self.broadcast(path)`.

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: Okay! I went with a simple if-else because the `strategy` flag only accepts two values and would error out if an invalid value is passed to it. But it's good to validate it here as well I think!

demo/lightning/checkpoint/multinode/fsspecfsdp.py
Yash9060: When str "gs://bucket-name" is converted to Path(object) it will get converted to Path(gs:/bucket-name). Since it sort of considers it as NFS path and drops `/`

I faced the same issue long time ago and hence we have this function https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/dataflux_pytorch/lightning/path_utils.py#L10 

So either what Abhi is doing is fine or he will have pass the path via the function linked above. 

But if yeah agreed if he is going to use string path then above code _might be_ unnecessary

demo/lightning/checkpoint/multinode/fsspecfsdp.py
Yash9060: full checkpoint is being returned in case of no sharding, 

if it's sharded it returns just metadata https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/strategies/fsdp.py#L631

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
Yash9060: Oh perfect. Thanks for testing it out!

demo/lightning/checkpoint/multinode/fsspecfsdp.py
abhibyreddi: That's right! But for sharded checkpoints it [returns just the metadata](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/strategies/fsdp.py#L631)

dataflux_pytorch/benchmark/checkpointing/multinode/fsspecfsdp.py
Yash9060: Yeah its kind of how it works, and hence we initialize strategy again in every iteration https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/dataflux_pytorch/benchmark/checkpointing/multinode/train.py#L84 

The reason why I am initializing reader/writer outside of save_checkpoint / load_checkpoint is because it is not fair to count the time it takes to initialize as part of save_checkpoint/load_checkpoint

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
jdnurme: You can set one of them as "default so it won't error out if omitted.

demo/lightning/checkpoint/multinode/fsspecfsdp.py
abhibyreddi: Replied below. This could just be `self.broadcast(filepath)`.

demo/lightning/checkpoint/multinode/fsspecfsdp.py
jdnurme: Ok, is this broadcast operation meant to be done on all nodes or just the rank 0 node?

Yash9060 (approved):
Great work Abhi !

demo/lightning/checkpoint/multinode/fsspecfsdp.py
awonak: Or perhaps leave your code as is and we can refactor at a later time.

jdnurme (approved):

demo/lightning/checkpoint/multinode/fsspecfsdp.py
abhibyreddi: This looks much better, Adam! You're right that `FSSpecFSDPStrategy` has a lot of duplication with the `DatafluxFSDPStrategy`. In fact, the former is just a copy-paste of the latter. I replaced Dataflux specific code with that of FSSpec. It took a lot of trial and error to figure out what changes to make and where.

I will file a follow-up task and work on making the changes you suggested here. Thanks for the example implementation!

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: Yes! I set the default value to the one that corresponds to `DatafluxFSDPStrategy`

awonak (approved):

PR #146, opened 2024-10-15
Correcting the code for simulated benchmarks

Yash9060:
Fixing the load_checkpoint code. Root cause was that load_checkpoint needs to know the model format (number of layers, size of layers, num_nodes etc etc) and this all format should be known to state_dict and hence empty_dict although would read data, it would be discarded. 

I have also added some code which will help with debugging (This debug code can write model & checkpoint in human readable format). Right now this code is in demo but I will later move it to benchmarks with some improvements(since benchmark code runs on GKE and hence we will need to upload this txt to GCS bucket). Right now the priority is getting numbers. 

Sample Run:
```
Completed iteration 2 for load ...
Time taken to save checkpoint: 4.0865 seconds
Time taken to load checkpoint: 17.7235 seconds
Size of distributed tensors (rank 0): 337.21 MB
Total size of all tensors: 1.65 GB
```


awonak (approved):

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -15,24 +15,120 @@
...
-from dataflux_pytorch.lightning.gcs_filesystem import GCSDistributedWriter, GCSDistributedReader
 
+from dataflux_pytorch.lightning.gcs_filesystem import GCSDistributedReader, GCSDistributedWriter
+
+# Constants for distributed setup
 MASTER_ADDR = 'localhost'
 MASTER_PORT = '12355'
 
 
+def write_model_structure(file: TextIO, model: nn.Module, indent: str = '') -> None:
```
jdnurme, line 38: I would specifiy `_to_file` in the name to disambiguate from writing to the actual model itself.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -15,24 +15,120 @@
...
+
+            if hasattr(model, 'dummy_tensors'):
+                f.write("\nDummy Tensors:\n")
+                for i, tensor in enumerate(model.dummy_tensors):
+                    f.write(f"dummy_tensor_{i}:\n")
+                    f.write(f"  Shape: {tensor.shape}\n")
+                    f.write(f"  Values: {tensor}\n")
+
+
+def write_state_dict(state_dict: Dict[str, torch.Tensor], filename: str) -> None:
```
jdnurme, line 78: add `_to_file` like above

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -15,24 +15,120 @@
...
+        f.write("State Dict:\n")
+        for key, value in state_dict.items():
+            f.write(f"{key}:\n")
+            f.write(f"  Shape: {value.shape}\n")
+            f.write(f"  Values: {value}\n")
+
+
+def format_size(size_bytes: int) -> str:
+    """Formats bytes into a human-readable size string."""
+    if size_bytes < 1024:
```
jdnurme, line 89: I would recommend setting 1024 to a constant called MB (or something similar) and reuse that. Reduces chance of accidentally altering one of the magic numbers.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -15,24 +15,120 @@
...
+                        default=100, help="Size of each layer.")
+    parser.add_argument("--clear-kernel-cache", action="store_true",
+                        default=False, help="Clear kernel cache.")
+    parser.add_argument("--sample-count", type=int, default=3,
+                        help="Number of samples for benchmarking.")
+    parser.add_argument("--padding-size", type=int, default=1000,
+                        help="Size of dummy tensors for padding.")
+    parser.add_argument("--world-size", type=int, required=True,
+                        help="Number of processes in the distributed setup.")
+    parser.add_argument("--debug", action="store_true",
```
jdnurme, line 125: include `default=False` for this flag.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -54,37 +150,39 @@ def save_checkpoint(self, checkpoint: Dict[str, torch.Tensor], filepath: str, st
...
                      storage_writer=self.writer)
 
-    def load_checkpoint(self, checkpoint_path: str) -> None:
+    def load_checkpoint(self, checkpoint_path: str, initial_state_dict: Dict) -> None:
         """
         Loads a model's state dictionary from a specified checkpoint file in GCS. torch.distributed.checkpoint.load contains the core logic of loading sharded model weights.
         You can find the source code for FSDP.load_checkpoint
         https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/fabric/strategies/fsdp.py#L519.
 
+        For torch.distributed.checkpoint.load to work properly the template_state_dict should have model format.
```
jdnurme, line 159: What does "model format" mean here? Match the dimensions of the model's input tensor?

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -235,7 +310,7 @@ def main():
     args = parse_args()
 
     mp.set_start_method('spawn')
-    mp.spawn(run_benchmark, args=(args.world_size, args.layer_size, args.project, args.ckpt_dir_path, args.padding_size, args.sample_count),
+    mp.spawn(run_benchmark, args=(args.world_size, args.layer_size, args.project, args.ckpt_dir_path, args.padding_size, args.sample_count, args.debug),
```
jdnurme, line 313: Based on this formatting I don't think you're VS code is setup with the correct formatter. We are using the vscode settings specified in the `.vscode` dir in the root of this project. Can you double check that those are configured and active? Once made active you should be able to just save and have it reformat automatically in one pass.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -176,53 +238,66 @@ def time_checkpoint_operation(benchmark_strategy: BenchmarkStrategy, distributed
             benchmark_strategy.save_checkpoint(
                 distributed_state_dict, filepath=checkpoint_path)
         elif operation == 'load':
-            benchmark_strategy.load_checkpoint(checkpoint_path=checkpoint_path)
+            benchmark_strategy.load_checkpoint(
+                checkpoint_path=checkpoint_path, initial_state_dict=template_state_dict)
         end_time = time.time()
         times.append(end_time - start_time)
         dist.barrier()  # Synchronize processes
```
jdnurme, line 245: Do we want the barrier call after the end time? Should it not be prior to end_time? Are we trying to get a timing print out for each node, or a time for the whole operation? If we want timing for each node individually we should specify what node we're running on in the print statement below.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -176,53 +238,66 @@ def time_checkpoint_operation(benchmark_strategy: BenchmarkStrategy, distributed
...
-    distributed_state_dict = {f"{key}_shard_{rank}": split_tensor(
-        tensor, world_size, rank) for key, tensor in full_state_dict.items()}
+    state_dict = model.state_dict()
+    for i, tensor in enumerate(model.dummy_tensors):
+        state_dict[f'dummy_tensor_{i}'] = tensor
+
+    if rank == 0 and debug:
+        print("Writing state dict before saving to file...")
+        write_state_dict(state_dict, "state_dict_before_save.txt")
+        print("Shapes before saving:", {
```
jdnurme, line 268: Is this printout massive? Do we want this?

jdnurme (approved):

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -95,61 +184,18 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-    """Splits a tensor into chunks for distributed processing."""
-    numel = tensor.numel()
-    split_size = numel // world_size
-    start_idx = rank * split_size
-    end_idx = start_idx + split_size if rank != world_size - 1 else numel
-    return tensor.view(-1)[start_idx:end_idx]
-
-
-def time_checkpoint_operation(benchmark_strategy: BenchmarkStrategy, distributed_state_dict: Dict[str, torch.Tensor], filepath: str, sample_count: int, operation: str) -> list:
+def time_checkpoint_operation(benchmark_strategy: BenchmarkStrategy, distributed_state_dict: Dict[str, torch.Tensor], filepath: str, sample_count: int, operation: str, model) -> list:
```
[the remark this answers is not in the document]
Yash9060: Done

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Done

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Done

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Done

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Correct.
Number of layers, size of tensors.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Not really massive, since I am only printing shape of each tensor. 
But it will be decent sized chunk as well. 
I think it is good to keep in the code (this will be later moved to benchmarks). So it can be used by devs with small model to check for correctness.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Just synced my settings to settings.json

Yash9060: Moved the barrier before end_time. Dont need it after since in for loop we already have dist.barrier() before start_time.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Done

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Just synced the formatting

PR #148, opened 2024-10-15
Set pytorch version to 2.3.1

abhibyreddi:
Pytorch v2.4.0 has some bugs that were causing checkpoint loads to fail. It has been verified locally that downgrading pytorch to v2.3.1 resolves those issues. 

This change has been tested locally by running `pip install .` and verifying torch version from python interpreter by running `torch.__version__`.


jdnurme (approved):

Yash9060 (approved):

abhibyreddi:
Update the wrong file.

MattIrv (approved):

PR #149, opened 2024-10-15
Multi-node checkpoint benchmark improvements

MattIrv:
Several minor fixes/improvements for the multi-node checkpoint benchmark:

- Update the README with some missing instructions
- Run as a module rather than directly with Python so that imports work correctly
- Update `WORLD_SIZE` env var to `NUM_NODES` because PyTorch / PyTorch Lightning appear to use `WORLD_SIZE` for other purposes
- Set `use_orig_params=False` when creating the strategy, since this avoids the time-consuming step of flattening the optimizer state dict
  - This happened on each node, so I'm hopeful that this gives us a better trend for throughput when scaling the number of nodes
  - This also required updating the model's `configure_optimizers` to use `self.trainer.model.parameters` instead of `self.model.parameters` as in the default (see https://github.com/Lightning-AI/pytorch-lightning/issues/17515)
- Make `NUM_DEVICES` a configurable environment variable
- Add the ability to configure the strategy with AdamW or SGD optimizer. AdamW appears to result in larger checkpoint files that are about 20% of the GPU memory utilization

- [x] Tests pass

### DIFF
--- a/demo/lightning/checkpoint/multinode/train.py
+++ b/demo/lightning/checkpoint/multinode/train.py
@@ -7,6 +7,7 @@ from typing import Generator
 
 import torch
 import torch.distributed
+import torch.optim
 from dataflux_core import user_agent
 from google.cloud import storage
 from lightning import Trainer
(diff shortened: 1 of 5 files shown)


demo/lightning/checkpoint/multinode/train.py
```diff
@@ -5,10 +5,9 @@
 from typing import Generator
 
 import torch
+import torch.distributed
```
Yash9060, line 8: It doesnt seem like newly added code uses torch.distributed ? 

Also it would much clearer (aligns with the examples I see) to use
` import torch.distributed as dist`

Yash9060 (approved):

abhibyreddi (approved):

PR #153, opened 2024-10-15
Skip saving checkpoints during training

abhibyreddi:
Currently checkpoints are saved during `trainer.fit` which is not needed because we manually save checkpoints by calling `trainer.save_checkpoint`

- [X] Tests pass - ran locally on a VM with 4 GPUs and on a GKE cluster with 2 nodes

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
@@ -101,7 +101,8 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
         model = DemoTransformer(vocab_size=dataset.vocab_size,
                                 nlayers=int(os.environ.get("NUM_LAYERS", 10)))
         new_ckpt_dir_path = os.path.join(ckpt_restore_path, f'ckpt_{i}.ckpt/')
-        strategy = get_strategy(args.strategy, model, new_ckpt_dir_path)
+        strategy = get_strategy(args.strategy, os.getenv("PROJECT"), model,
+                                new_ckpt_dir_path)
         trainer = Trainer(
             enable_checkpointing=False,
             default_root_dir=ckpt_dir_path,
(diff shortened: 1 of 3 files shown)


dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -72,29 +69,17 @@ def main(project: str,
...
-    checkpoint_callback = ModelCheckpoint(
-        save_top_k=1 if save_only_latest else -1,
-        every_n_train_steps=1,
-        filename="checkpoint-{epoch:02d}-{step:02d}",
-        enable_version_counter=True,
-    )
-
     strategy = get_strategy(args.strategy, model, ckpt_dir_path)
-    min_epochs_save = int(os.environ.get("MIN_EPOCHS_SAVE", 4))
-    max_epochs_save = int(os.environ.get("MAX_EPOCHS_SAVE", 5))
```
MattIrv, line 87: Can you update the readme to remove these as well? And possibly the deployment yaml file too if they're set there

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -59,10 +59,7 @@ def get_strategy(choice, model, ckpt_dir_path):
     return strategy
 
 
-def main(project: str,
```
MattIrv, line 62: nit: As much as possible it's probably better to try to leave os.environ calls inside of main or the `if __name__ == '__main__'` block, and pass parameters to functions like `get_strategy`. It's not a major difference, but it ensures they can be cleanly reused outside of this code if needed

MattIrv (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: These vars are not set in the deployment file. I renamed `MAX_STEPS_SAVE` and `MAX_STEPS_RESTORE` to `NUM_SAVE_CALLS` and `NUM_LOAD_CALLS` and updated the README. Do we really need to have these variables configurable?

awonak (approved):
Nice improvements!

I realized that we didn't add a note in the README for the different strategy command line args like `dataflux_fsdp` and `fsspec_fsdp`. Can you add those details here or in a follow up PR?

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
Yash9060: +1

Yash9060 (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: Done.

PR #156, opened 2024-10-17
Add an option to benchmark only save calls

abhibyreddi:
Tested locally on a VM with 4 GPUs with and without `--save_only`.

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
@@ -129,8 +129,8 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
             print(f"Loaded checkpoint from {new_ckpt_dir_path}.")
         load_checkpoint_times.append(end - start)
 
-    avg_load_time = statistics.mean(load_checkpoint_times)
     if torch.distributed.get_rank() == 0:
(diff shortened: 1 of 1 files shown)


dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -95,42 +96,46 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
...
-            strategy=strategy,
-            devices=os.environ.get("NUM_DEVICES", 'auto'),
-            num_nodes=num_nodes,
-        )
-        trainer.fit(model, dataloader, ckpt_path=new_ckpt_dir_path)
-        start = time.time()
-        trainer.strategy.load_checkpoint(new_ckpt_dir_path)
-        end = time.time()
+    if not args.save_only:
+        num_load_calls = int(os.environ.get("NUM_LOAD_CALLS", 3))
```
MattIrv, line 100: Could you not just set NUM_LOAD_CALLS to 0 to get effectively this same behavior?

abhibyreddi: Done! Kept the `save_only` flag because it's needed to skip computing mean load times. `statistics.mean()` expects a non-empty array.

abhibyreddi: Done.

PR #157, opened 2024-10-17
Implement a custom FSDP strategy for benchmarking loads from boot disk

abhibyreddi:
This is needed to be able to benchmark checkpoint saves/loads to/from boot disk in distributed environment

* When run with `--save_only`, lightning's FSDP strategy is used. Load calls are skipped.
* When run with `--load_only` newly added `LoadFromBootDiskFSDP` is used which saves with dataflux. All nodes then copy all the contents of the GCS bucket to boot disk and then load their checkpoints. Avg. time to save one checkpoint will be reported as `skipped`.
* Move all custom strategy classes to  `demo/lightning/checkpoint/multinode/strategies.py`

- [X] Tests pass - ran manually on a VM with GPUs and on a GKE cluster with 2 nodes
- [X] Appropriate changes to documentation are included in the PR - follow up PR will update the README


abhibyreddi:
Contents of `demo/lightning/checkpoint/multinode/strategies.py` are mostly reviewed already except for the `CustomFSDPStrategy` class which is introduced in this PR.

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -6,15 +6,15 @@
 import torch
 import torch.distributed
 from lightning import Trainer
-from lightning.pytorch.callbacks import ModelCheckpoint
 from lightning.pytorch.demos import WikiText2
-from lightning.pytorch.strategies import FSDPStrategy
 import torch.distributed
 from torch.utils.data import DataLoader
+from lightning.pytorch.strategies import FSDPStrategy
+from google.cloud import storage
```
MattIrv, line 13: Do you have the workspace recommended vscode `isort` extension installed and activated? Not sure if these are in the right order

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -97,14 +117,24 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
     avg_save_time = (end - start) / num_save_calls
     num_load_calls = int(os.environ.get("NUM_LOAD_CALLS", 3))
     load_checkpoint_times = []
+    if args.save_only:
+        print("Skipping loads because you set --save_only")
+        num_load_calls = 0
+        load_checkpoint_times = [0]
+    if args.strategy == FSDP_STRATEGY and args.load_only:
```
MattIrv, line 124: Does there need to be some other behavior if you set load_only without FSDP_STRATEGY?

demo/lightning/checkpoint/multinode/strategies.py
```diff
@@ -0,0 +1,239 @@
...
+        # Load metadata (anything not a module or optimizer)
+        new_path = os.path.join(checkpoint_path, _METADATA_FILENAME)
+        metadata = None
+        with self.reader.fs.create_stream(path=new_path,
+                                          mode='rb') as metadata_file:
+            metadata = torch.load(metadata_file)
+        return metadata
+
+
+class CustomFSDPStrategy(FSDPStrategy):
```
MattIrv, line 220: Could you add a doc comment explaining how this is different from the other strategies? Is there a more descriptive name we could use here?

MattIrv (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: I do not! I will install it right away.

demo/lightning/checkpoint/multinode/strategies.py
abhibyreddi: Renamed it to `LoadFromBootDiskFSDP` and added a doc comment. Does that look good?

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: Done!

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: Added a function named `validate` that throws an error when things aren't set as expected. Update this conditional accordingly. Could you take another look?

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -97,14 +115,24 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
...
+        print(f"Copying contents of {ckpt_dir_path} to {ckpt_restore_path}")
+        copy_bucket_to_local(ckpt_dir_path.removeprefix("gs://"),
+                             os.path.dirname(ckpt_restore_path))
+        avg_save_time = 0
```
jdnurme, line 141: You should probably gate the printout for save/load with the same flag below so the output isn't confusing with the 0 seconds.

jdnurme (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -97,14 +130,24 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
     avg_save_time = (end - start) / num_save_calls
     num_load_calls = int(os.environ.get("NUM_LOAD_CALLS", 3))
     load_checkpoint_times = []
+    if args.save_only:
+        print("Skipping loads because you set --save_only")
+        num_load_calls = 0
+        load_checkpoint_times = [0]
+    elif args.load_only:
+        print(f"Copying contents of {ckpt_dir_path} to {ckpt_restore_path}")
+        copy_bucket_to_local(ckpt_dir_path.removeprefix("gs://"),
```
jdnurme, line 139: It may not matter, but you already should have a storage client configured within the model at this point right? Not sure if it matters whether or not we open two though.

jdnurme (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: How about I just print "skipped" instead of "0 seconds"

abhibyreddi: Done.

MattIrv (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
abhibyreddi: https://screenshot.googleplex.com/6FjQKCzc7wpjbNu

PR #161, opened 2024-10-21
Correcting load_checkpoint for simulated benchmarks. 

Yash9060:
Code changes for simulated benchmarking, load_checkpoint.
Updated doc strings/ comments, and removed unnecessary sections. 

Will update documentation in follow up PR.

- [X] Tests pass
- [] Appropriate changes to documentation are included in the PR


dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
```diff
@@ -49,56 +44,64 @@ def init_processes() -> int:
...
+    layer_size = int(os.getenv("LAYER_SIZE"))
     project = os.getenv("PROJECT")
-    path = os.getenv("CKPT_DIR_PATH")
-    sample_size = int(os.getenv("SAMPLE_COUNT", 3))
+    ckpt_dir_path = os.getenv("CKPT_DIR_PATH")
+    sample_count = int(os.getenv("SAMPLE_COUNT", 3))
     padding_size = int(os.getenv("PADDING_SIZE", 4000))
-    main(world_size, model_size,
-         project, path, padding_size, sample_size)
+    main(world_size, layer_size, project, ckpt_dir_path, padding_size,
```
jdnurme, line 106: main() shouldn't really be taking arguments. We should move these environment variable grabs into a setup function and call it from main.

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
```diff
@@ -49,56 +44,64 @@ def init_processes() -> int:
...
-        benchmark_strategy, distributed_state_dict, filepath, sample_size, 'load')
+        project=project,
+        path=filepath,
+    )
+
+    state_dict = dict()
+    for i in range(padding_size):
+        if i % world_size == rank:
+            state_dict[f'dummy_tensor_{i}'] = torch.randn(layer_size, 1000)
+    dist.barrier()
```
jdnurme, line 66: Perhaps add a small comment above the barrier call explaining we need to wait for all nodes to finish populating with dummy tensors before moving forward.

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
```diff
@@ -49,56 +44,64 @@ def init_processes() -> int:
...
-        print(
-            f"Total size of all tensors (rank {rank}): {format_size(total_distributed_size_bytes * world_size)}")
+        print(f"Time taken to save checkpoint:\
+                {statistics.mean(save_checkpoint_times):.4f} seconds")
+        print(f"Time taken to load checkpoint:\
+                 {statistics.mean(load_checkpoint_times):.4f} seconds")
+
+        tensor_size_per_instance = 1000 * layer_size * state_dict[
+            f'dummy_tensor_0'].element_size()
+        total_tensor_count = padding_size // world_size  # Total tensors per rank
```
jdnurme, line 88: probably should name this `tensors_per_rank` since it's not the total.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -268,12 +198,14 @@ def time_checkpoint_operation(benchmark_strategy: BenchmarkStrategy,
...
+    # According to `create_default_local_load_plan` https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/default_planner.py#L343
+    # each key will be read only once from the state_dict, hence assigning different names to different tensor will force the load function to only read
+    # tensor shard corresponding to given node.
+    for i in range(tensor_count):
+        if i % world_size == rank:
+            template_state_dict[f'dummy_tensor_{i}'] = torch.empty(
+                tensor_size, 1000)
     for i in range(sample_count):
         checkpoint_path = os.path.join(filepath, f'checkpoints/ckpt_{i}.ckpt')
         dist.barrier()
```
jdnurme, line 211: Should this dist.barrier() call be between the two for loops? Do we need it inside the loop given we already have a barrier call below after the save/load combo?

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -322,24 +251,29 @@ def run_benchmark(rank, world_size: int, layer_size: int, project: str,
...
         print(f"Time taken to save checkpoint:\
                 {statistics.mean(save_checkpoint_times):.4f} seconds")
         print(f"Time taken to load checkpoint:\
                  {statistics.mean(load_checkpoint_times):.4f} seconds")
-        total_distributed_size_bytes = sum(
-            get_tensor_size_bytes(tensor) for tensor in state_dict.values())
+
+        tensor_size_per_instance = 1000 * layer_size * state_dict[
+            f'dummy_tensor_0'].element_size()
+        total_tensor_count = padding_size // world_size  # Total tensors per rank
```
jdnurme, line 271: Should call this var `tensors_per_rank` or equivalent.

jdnurme (approved):

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
Yash9060: We already have a different version under demo for kokoro integration https://github.com/GoogleCloudPlatform/dataflux-pytorch/tree/main/demo/lightning/checkpoint/simulated , so I guess this can remain as it is ? 

Offcourse when someone takes on b/2F367371602, they can decide on long term strategy on how to combine code between benchmarks and demo.

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
Yash9060: Done

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
Yash9060: Done

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Not really, I want all the nodes to start and end the save/load checkpoint at the same time, so that we can see if number of concurrent requests to GCS causes any regressions in the throughput. 

I dont see any problem if dist.barrier() is not present between the for loop, since the names of tensors are different.

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
```diff
@@ -49,56 +44,66 @@ def init_processes() -> int:
...
-
-    save_checkpoint_times = time_checkpoint_operation(
-        benchmark_strategy, distributed_state_dict, filepath, sample_size, 'save')
-    load_checkpoint_times = time_checkpoint_operation(
-        benchmark_strategy, distributed_state_dict, filepath, sample_size, 'load')
+        project=project,
+        path=filepath,
+    )
+
+    state_dict = dict()
```
MattIrv, line 62: A comment explaining this block might be helpful

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
```diff
@@ -49,56 +44,66 @@ def init_processes() -> int:
...
+                                                      padding_size, layer_size)
+
+    load_checkpoint_times = time_checkpoint_operation(benchmark_strategy,
+                                                      state_dict, filepath,
+                                                      sample_count, 'load',
+                                                      rank, world_size,
+                                                      padding_size, layer_size)
 
     if rank == 0:
-        print("######################")
```
MattIrv, line 80: Did you mean to get rid of this print at the start?

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
```diff
@@ -49,56 +44,66 @@ def init_processes() -> int:
...
         print("######################")
 
     cleanup()
 
 
 if __name__ == "__main__":
     world_size = int(os.getenv("WORLD_SIZE"))
-    model_size = int(os.getenv("NUM_LAYERS"))
+    layer_size = int(os.getenv("LAYER_SIZE"))
+    layer_size = int(os.getenv("LAYER_SIZE"))
```
MattIrv, line 103: Duplicated code, is one of them supposed to be something else?

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
MattIrv: Not sure if there is any standard or conventional style in Python for this, but my sense is that main usually should not be grabbing environment variables (i.e. so that it could be called as library code from somewhere else if you wanted to) and this is a good place to do it.

demo/lightning/checkpoint/simulated/multiprocessing_train.py
MattIrv, line 1: Is there any way to reduce the code duplication between this and the benchmark code?

Honestly I'm not sure if it really makes sense to have a simulated "demo" codebase. The only reason we want to simulate it is for benchmarking.

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
Yash9060: Nope I just feel it helps capture the entire block between two set of ###

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
Yash9060: Removed (messed up during the merge rebase)

dataflux_pytorch/benchmark/checkpointing/simulated/multi_node_benchmark.py
jdnurme: If we want to keep the env var args here thats ok, but we should still not pass individual args in main. Instead we can pass an argv single param containing a sequence like here: https://engdoc.corp.google.com/eng/doc/devguide/py/style/index.md?cl=head

Yash9060: I can do something like this https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/demo/lightning/checkpoint/simulated/multiprocessing_train.py#L356 

main doesnt take any args, but it parses env var and then calls run_benchmark. Sounds good ? @jdnurme @MattIrv

jdnurme: Yeah that looks fine to me. Longterm we can also create some sort of config object for this that we can share across all demo code, but that's definitely out of scope for this. We will eventually need to sort out the mix of command line vs env var stuff we have.

Yash9060: Done

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Yeah there is significant overlap between demo & benchmark code. There are some decisions that need to be made on how to structure the code between benchmark and demo but its part of b/367371602

PR #164, opened 2024-10-22
Adding option to use FSspec with simulated benchmarks

Yash9060:
Add option to use FSSpec with simulated benchmarks. 

- [X] Tests pass

### DIFF
--- a/demo/lightning/checkpoint/simulated/multiprocessing_train.py
+++ b/demo/lightning/checkpoint/simulated/multiprocessing_train.py
@@ -27,6 +27,7 @@ import torch.nn as nn
 from dataflux_pytorch.lightning.gcs_filesystem import (GCSDistributedReader,
                                                        GCSDistributedWriter)
 from lightning.pytorch.strategies import FSDPStrategy
+from torch.distributed.checkpoint import _fsspec_filesystem as FF
 
 # Constants for distributed setup
 MASTER_ADDR = 'localhost'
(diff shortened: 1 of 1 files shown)


demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -103,15 +104,27 @@ def parse_args() -> argparse.Namespace:
...
 
 class BenchmarkStrategy(FSDPStrategy):
 
-    def __init__(self, project: str, path: str, **kwargs):
+    def __init__(self, project: str, path: str, use_fsspec: bool, **kwargs):
         super().__init__(**kwargs)
-        self.writer = GCSDistributedWriter(path, project, None)
-        self.reader = GCSDistributedReader(path, project, None)
+        if use_fsspec:
+            self.reader = FF.FsspecReader(path)
```
jdnurme, line 123: Is the provided path different for Fsspec vs GCSDist given that we don't pass it the project? or same path?

jdnurme (approved):

MattIrv (approved):

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Its the same path. Reader/Writer must be doing some sort of auth in lower layers. And since bucket name is unique universally, I dont think it needs project name at all.

PR #165, opened 2024-10-24
Add Llama2 7B FSDP demo

abhibyreddi:
Demo code performs data loading, listing, and checkpoint saving with Dataflux. Follow up PR will add support for loading the saved checkpoints with Dataflux. 

- [X] Tests pass - tested manually on a GCE VM with GPUs


abhibyreddi:
Matt & JD, all the files introduced in this PR are borrowed from [here](https://github.com/Lightning-AI/lit-llama/tree/main/lit_llama). I've called out the changes _I_ made in the README file. 

I tested this by training the 7B model on the sample dataset (~`24GB`) for `500` iterations. I also added the output I got when I prompted the generated model (it's very bad but it works!)

demo/llama/dataset.py
MattIrv, line 1: Instead of copying this code and other lit-llama code directly into our codebase, we should either bring it into a third-party folder (there are instructions internally for how to structure this for GitHub, but let me know if you can't find them) or maybe a better choice would be to include lit-llama as a git submodule in this folder like we do for dataflux-client-python and maxtext repos elsewhere

demo/llama/README.md
```diff
@@ -0,0 +1,65 @@
+The code in this directory trains the Llama 7B model on [Huggingface's RedPajama dataset](https://huggingface.co/datasets/togethercomputer/RedPajama-Data-1T-Sample/tree/main). All the code in this directory is borrowed from [Lightning AI's lit-llama repo](https://github.com/Lightning-AI/lit-llama). Changes have been made in appropriate places to fetch the data from GCS instead of reading from disk.
+
+This demo has been tested on a GCE instance with `2` Nvidia `H100` GPUs.
+
```
MattIrv, line 4: Let's scope the demo to just fsdp checkpointing, if possible.

abhibyreddi: Made changes to support saving checkpoints with Dataflux. Updated README with details. Could you take another look?

abhibyreddi:
@MattIrv, @jdnurme , @Yash9060 want to call your attention to  `demo/llama/strategies.py`. The `DatafluxFSDPStrategy` class there is similar to the class with the same name in `demo/lightning/checkpoint/multinode/strategies.py`.

The one introduced in this PR inherits `lightning.fabric.strategies.FSDPStrategy`. The existing one inherits `lightning.pytorch.strategies.FSDPStrategy`. They are similar but not the same.

demo/lightning/checkpoint/multinode/strategies.py
```diff
@@ -7,7 +7,11 @@
 import torch
 import torch.distributed as dist
 from dataflux_core import user_agent
+from dataflux_pytorch.lightning import DatafluxLightningCheckpoint
+from dataflux_pytorch.lightning.gcs_filesystem import (GCSDistributedReader,
+                                                       GCSDistributedWriter)
 from google.cloud import storage
+from lightning.fabric.utilities.types import _PATH
```
Yash9060, line 14: Dont see _PATH being used here ? Do we need it ?

demo/llama/README.md
```diff
@@ -0,0 +1,67 @@
...
+### Download the data
+
+Follow the instructions [here](https://github.com/Lightning-AI/lit-llama/blob/main/howto/train_redpajama.md#prepare-redpajama-for-training) to download the dataset. Upload the directory with `.bin` files generated by the pre-processing script to your GCS bucket.
```
Yash9060, line 7: nit: Do you think it would be useful add a command or 2 to help customers move data to GCS. ?

demo/llama/strategies.py
```diff
@@ -0,0 +1,241 @@
...
+        path = Path(self.broadcast(path))
+        writer = GCSDistributedWriter(path, self.storage_client.project,
+                                      self.storage_client)
+
+        start_time = time.time()
+        if self.use_async:
+            self._async_save(converted_state, path, writer)
+        else:
+            self._save(converted_state, path, writer)
+        duration_ms = (time.time() - start_time) / 1000
```
Yash9060, line 104: ANy specific reason why we need to have time in millisecond ? I think everywhere else in the codebase we use seconds ?

demo/llama/strategies.py
```diff
@@ -0,0 +1,241 @@
...
+        self.use_async = use_async
+        self.checkpoint_group = None
+        self._checkpoint_future = None
+
+        if self.use_async:
+            default_ranks = list(range(dist.get_world_size()))
+            self.checkpoint_group = dist.new_group(
+                default_ranks, backend=self.process_group_backend)
+
+    def save_checkpoint(
```
Yash9060, line 46: Add a link to save_checkpoint source code ? (Also I think this is coming from lightning fabric, any reason why we are using lightning fabric save_checkpoint instead of simple lightning checkpoint ?)

demo/llama/train.py
```diff
@@ -0,0 +1,386 @@
+# Copyright Lightning AI. Licensed under the Apache License 2.0, see LICENSE file.
+
```
Yash9060, line 2: Add source code link ?

demo/lightning/checkpoint/multinode/strategies.py
abhibyreddi: Got the wrong file. Removed the import.

demo/llama/README.md
abhibyreddi: Done.

demo/llama/train.py
abhibyreddi: Done.

demo/llama/README.md
MattIrv: The readme doesn't appear to cover checkpointing at all. Can you update it?

I think the training portion is not that interesting from readme purposes, since it doesn't involve Dataflux at all (does it?)

demo/llama/dataset.py
MattIrv, line 1: This appears to mostly be a copy of https://github.com/Lightning-AI/lit-llama/blob/main/lit_llama/packed_dataset.py. Is there some way to only override the pieces we need from that code without reimplementing everything?

This might apply for strategies.py and train.py here too

demo/llama/strategies.py
MattIrv, line 1: Is there a way to avoid adding yet another reimplementation of these strategies and reuse the existing ones?

demo/llama/strategies.py
abhibyreddi: lit-llama's training code does not use `lightning.Trainer` and `fabric` takes only custom strategies that inherit one of the classes defined in `lightning.fabric.strategies`. Another option I had was to re-write the training code to use lightning's `LightningModule`. This seemed simpler to do.

demo/llama/train.py
```diff
@@ -0,0 +1,389 @@
...
+    model, optimizer = fabric.setup(model, optimizer)
+
+    process_batch_size = batch_size // devices
+    gradient_accumulation_iters = process_batch_size // micro_batch_size
+
+    train(fabric, model, optimizer, train_dataloader, val_dataloader,
+          gradient_accumulation_iters, devices)
+
+
+def train(
```
MattIrv, line 136: Can the underlying implementation be reused here? It looks like there might not be any changes

demo/llama/train.py
```diff
@@ -0,0 +1,389 @@
...
+        if not is_accumulating:
+            tokens = 0
+            step_time = 0.0
+
+        if iter_num > max_iters:
+            break
+
+
+@torch.no_grad()
+def validate(fabric: L.Fabric, model: torch.nn.Module,
```
MattIrv, line 245: Same comment here and for several of the functions below, does this really need to be reimplemented?

MattIrv (changes requested):

demo/llama/strategies.py
abhibyreddi: This has been copied over from `demo/lightning/checkpoint/multinode/strategies.py`.

demo/llama/README.md
abhibyreddi: I added a note in the "Run the pre-training script" section about implementing a custom strategy that makes it possible to save checkpoints using Dataflux. What other details would be good to mention here?

PR #166, opened 2024-10-28
delete and reconstruct client

akansha1812:
When multiprocess is called, client will be deleted and recreated for each process . This will resolve the pickling error when a client is passed by a user for Mac and Windows OS. Along with this a client will not be constructed every time.  Reduce storage client information and reconstruct the client using this information.

referred: https://docs.python.org/3/library/pickle.html#handling-stateful-objects

- [ ] Tests pass

### DIFF
--- a/dataflux_pytorch/dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/dataflux_mapstyle_dataset.py
@@ -257,7 +257,6 @@ class DataFluxMapStyleDataset(data.Dataset):
         state = self.__dict__.copy()
         # Remove the unpicklable entries.
         del state['storage_client']
-        print("get_state: client deleted")
         return state
 
     def __setstate__(self, state):
@@ -266,4 +265,3 @@ class DataFluxMapStyleDataset(data.Dataset):
         # Restore the previously opened file's state. To do so, we need to
         # reopen it and read from it until the line count is restored.
         self.storage_client = self._unpickle_client()
-        print("set_state: client created")
--- a/dataflux_pytorch/dataflux_mapstyle_dataset.py
+++ b/dataflux_pytorch/dataflux_mapstyle_dataset.py
@@ -138,6 +138,9 @@ class DataFluxMapStyleDataset(data.Dataset):
         self.bucket_name = bucket_name
         self.data_format_fn = data_format_fn
         self.config = config
+        if self.storage_client is None:
+            self.storage_client = storage.Client(project=self.project_name)
+        self._pickle_client()
         # If composed download is enabled and a storage_client was provided,
         # check if the client has permissions to create and delete the
         # composed object.
@@ -165,7 +168,7 @@ class DataFluxMapStyleDataset(data.Dataset):
     def __getitem__(self, idx):
         return self.data_format_fn(
             dataflux_core.download.download_single(
-                storage_client=self._get_or_create_storage_client(),
+                storage_client=self.storage_client,
                 bucket_name=self.bucket_name,
                 object_name=self.objects[idx][0],
                 retry_config=self.config.download_retry_config,
@@ -178,7 +181,7 @@ class DataFluxMapStyleDataset(data.Dataset):
                 project_name=self.project_name,
                 bucket_name=self.bucket_name,
                 objects=[self.objects[idx] for idx in indices],
-                storage_client=self._get_or_create_storage_client(),
+                storage_client=self.storage_client,
                 dataflux_download_optimization_params=self.
                 dataflux_download_optimization_params,
                 threads=self.config.threads_per_process,
@@ -201,8 +204,8 @@ class DataFluxMapStyleDataset(data.Dataset):
                     retry_config=self.config.list_retry_config,
                 )
                 # If the dataset was not initialized with an storage_client, ensure that we do not attach a client to the lister to avoid pickling errors (#58).
-                if self.storage_client is not None:
-                    lister.client = self.storage_client
+                # if self.storage_client is not None:
+                #     lister.client = self.storage_client
                 listed_objects = lister.run()
 
             except Exception as e:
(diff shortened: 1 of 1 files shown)


dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -232,3 +234,18 @@ def _get_or_create_storage_client(self):
...
+        state = self.__dict__.copy()
+        # Remove the unpicklable entries.
+        del state['storage_client']
+        return state
+
+    def __setstate__(self, state):
+        # Restore instance attributes.
+        self.__dict__.update(state)
+        # Create the storage client.
+        self.storage_client = storage.Client(project=self.project_name)
```
jdnurme, line 251: This looks like a great change! Did you run any real workload tests to see what sort of time improvements we see?

jdnurme (approved):

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -165,7 +167,7 @@ def __len__(self):
     def __getitem__(self, idx):
         return self.data_format_fn(
             dataflux_core.download.download_single(
-                storage_client=self._get_or_create_storage_client(),
```
awonak, line 168: I believe all references to this method have been removed, so we can remove the method too.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -435,6 +434,69 @@ def test_init_with_perm(self):
...
+        # Act.
+        ds.__setstate__(state)
+
+        # Assert.
+        self.assertIn(
+            "storage_client",
+            ds.__dict__,
+            f"Key 'storage_client' should exist in dataflux_mapstyle_dataset instance",
+        )
+
```
awonak, line 499: optionally, you can further verify that the value of the "storage_client" key is the expected object type: 

```suggestion
    self.assertIsInstance(ds.__dict__['storage_client'], storage.Client)
```

awonak (approved):
Nice! This is a much better approach to handling the storage_client.

dataflux_pytorch/dataflux_mapstyle_dataset.py
```diff
@@ -201,8 +203,8 @@ def _list_GCS_blobs_with_retry(self):
                     retry_config=self.config.list_retry_config,
                 )
                 # If the dataset was not initialized with an storage_client, ensure that we do not attach a client to the lister to avoid pickling errors (#58).
-                if self.storage_client is not None:
-                    lister.client = self.storage_client
+                # if self.storage_client is not None:
```
jdnurme, line 206: These should either be deleted or not-commented out if you intend to keep them.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -337,8 +332,14 @@ def data_format_fn(content):
             retry_config=dataflux_mapstyle_dataset.MODIFIED_RETRY,
         )
 
-    def test_init_sets_user_agent(self):
+    @mock.patch("dataflux_pytorch.dataflux_mapstyle_dataset.dataflux_core")
```
jdnurme, line 335: It may not matter in this instance, but you don't have to mock the whole package out if you want to be more precise. If you namespace into the module (e.g. `fast_list.ListingController`) you can specifically mock out only that portion of dataflux core without mocking everything else. Generally speaking we want to mock the smallest amount of code necessary.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -353,26 +354,9 @@ def test_init_sets_user_agent(self):
...
-                    bucket_name=self.bucket_name,
-                    config=config,
-                    data_format_fn=self.data_format_fn,
-                    storage_client=client,
-                )
-
     @mock.patch("dataflux_pytorch.dataflux_mapstyle_dataset.dataflux_core")
     def test_list_GCS_blobs_with_spawn_multiprocess(self, mock_dataflux_core):
-        """Tests the _list_GCS_blobs_with_retry doesn't initialize client before calling dataflux_core.fast_list.ListingController when multiprcessing start method is spawn."""
+        """Tests the _list_GCS_blobs_with_retry initializes client before calling dataflux_core.fast_list.ListingController when multiprcessing start method is spawn."""
```
jdnurme, line 359: nit: maybe say "start method is set to spawn" so it reads a bit clearer.

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -414,9 +402,15 @@ def test_init_without_perm(self):
                 storage_client=client,
             )
 
-    def test_init_with_perm(self):
+    @mock.patch("dataflux_pytorch.dataflux_mapstyle_dataset.dataflux_core")
+    def test_init_with_perm(self, mock_dataflux_core):
         """Tests that the compose download is not disabled when create and delete permissions exists."""
         # Arrange.
+        # Arrange.
```
jdnurme, line 409: What does this comment mean?

awonak (approved):
LGTM

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -435,6 +429,69 @@ def test_init_with_perm(self):
             f"got max_composite_object_size for compose download{ds.config.max_composite_object_size}, want {want_size}",
         )
 
+    @mock.patch("dataflux_pytorch.dataflux_mapstyle_dataset.dataflux_core")
+    def test_getstate(self, mock_dataflux_core):
+        """Tests that the dataset.__getitems__ method returns the list of the correct downloaded object content."""
+        # Arrange.
```
jdnurme, line 435: again maybe we should be a bit more clear on what we mean by Arrange here. Maybe "setup mocks for testing" or something could be clearer?

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -435,6 +429,69 @@ def test_init_with_perm(self):
...
+        mock_listing_controller = mock.Mock()
```
jdnurme, line 436: If you want it to be more specific, you can actually pass an object like a listing controller as the argument to `mock.Mock()` and it will auto-mock in the format of that object (see the "spec" argument [here](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock))

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
```diff
@@ -435,6 +429,69 @@ def test_init_with_perm(self):
...
+            states,
+            want_state,
+            f"got dataflux_mapstyle_dataset params {states}, want {want_state}",
+        )
+
+    @mock.patch("dataflux_pytorch.dataflux_mapstyle_dataset.dataflux_core")
+    def test_setstate(self, mock_dataflux_core):
+        """Tests that the dataset.__getitems__ method returns the list of the correct downloaded object content."""
+        # Arrange.
+        mock_listing_controller = mock.Mock()
```
jdnurme, line 465: if this little chunk of code is the same everywhere you could move it to a test helper function to reduce repeat code.

jdnurme (approved):

dataflux_pytorch/tests/test_dataflux_mapstyle_dataset.py
akansha1812: I have changed the namespace to include fast_list.ListingController which should mock ListingController.
Updated it for all test cases where the same fast_list.ListingController is mocked.

PR #171, opened 2024-10-31
add simple llama load benchmark and results

jdnurme:
Fixes #<issue_number_goes_here>

- [x] Tests pass

### DIFF
--- a/dataflux_pytorch/benchmark/llama-70b/README.md
+++ b/dataflux_pytorch/benchmark/llama-70b/README.md
@@ -1,0 +1,20 @@
+# Load Benchmark for Llama-70b
+
+This benchmark is designed to give an approximation of load times for the [Llama-70b](https://huggingface.co/meta-llama/Llama-2-70b-hf/tree/main) model. This benchmark does not account for boot time of the virtual machine, or time to load the model into VRAM. This is a pure load of the model into memory from a GCS bucket. For this benchmark, the model is saved in 15 distinct `.bin` files. This benchmark iterates sequentially through each file, calling `torch.load()` on the dataflux checkpoint reader.
+
+## Benchmark Results
+
+This benchmark was performed on an [n2-standard-64](https://cloud.google.com/compute/docs/general-purpose-machines#n2_machine_types) VM with 256GB of memory in a single-node configuration. The bucket under test was co-located with the VM in the us-central1 region.
+
+| Run | Time to Load Model (seconds) |
+| --- | --- |
+|  1  | 651 |
+|  2  | 646 |
+|  3  | 652 |
+|  4  | 712 |
+|  5  | 705 |
+|  6  | 682 |
+|  7  | 721 |
+|  8  | 703 |
+
+Average across 8 benchmark runs: 684 seconds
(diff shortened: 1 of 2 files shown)


dataflux_pytorch/benchmark/llama-70b/README.md
```diff
@@ -0,0 +1,20 @@
...
+|  1  | 651 |
+|  2  | 646 |
+|  3  | 652 |
+|  4  | 712 |
+|  5  | 705 |
+|  6  | 682 |
+|  7  | 721 |
+|  8  | 703 |
+
+Average across 8 benchmark runs: 684 seconds
```
MattIrv, line 20: I would recommend adding the corresponding calculated throughput here too

dataflux_pytorch/benchmark/llama-70b/load.py
```diff
@@ -0,0 +1,37 @@
+import time
+
+import torch
+
+from dataflux_pytorch import dataflux_checkpoint
+
+start_time = time.time()
+ckpt = dataflux_checkpoint.DatafluxCheckpoint(project_name="<PROJECT_NAME>",
+                                              bucket_name="<BUCKET_NAME")
```
MattIrv, line 9: missing a bracket in this placeholder

Might be better to use flags or environment variables like our other demos/benchmarks

MattIrv (approved):

dataflux_pytorch/benchmark/llama-70b/load.py
Yash9060: +1

dataflux_pytorch/benchmark/llama-70b/load.py
```diff
@@ -0,0 +1,37 @@
...
+CKPT_PATH = "llama2/llama2-70b-hf/"
+LOAD_COUNT = 10
+files = [
```
Yash9060, line 12: nit: we can simplify generation of file names ?

dataflux_pytorch/benchmark/llama-70b/load.py
```diff
@@ -0,0 +1,37 @@
...
+    'pytorch_model-00015-of-00015.bin'
+]
+
+times = 0
+for i in range(LOAD_COUNT):
+    print(f"--------- RUN {i} ----------")
+    print("Performing read...")
+    for f in files:
+        with ckpt.reader(CKPT_PATH + f) as reader:
+            read_state_dict = torch.load(reader)
```
Yash9060, line 29: If we are just benchmarking read code should we not have start_time just before the loop starts instead of including it at the top of the file (although I do understand it wont result in any major changes in terms of final result)

dataflux_pytorch/benchmark/llama-70b/load.py
jdnurme: I just want to leave it explicit since this matches the exact format of the files from huggingface. There are multiple sets of files (such as the safe-tensor set)

dataflux_pytorch/benchmark/llama-70b/load.py
jdnurme: Good catch. The bench was actually run that way to begin with, my formatting got messed up and it was put to the top when I was cleaning it up for github.

Yash9060 (approved):

PR #173, opened 2024-11-04
Remove model and path arguments from FSDP strategy constructors

MattIrv:
This makes the constructor signature of our FSDP strategies more consistent with the base one, and ensures that the strategies can handle multiple different paths at runtime. `self.model` on the strategy is still set by the Lightning trainer when calling trainer.fit.

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
@@ -5,15 +5,16 @@ import time
 
 import torch
 import torch.distributed
-from demo.lightning.checkpoint.multinode.strategies import (
-    DatafluxFSDPStrategy, FSSpecFSDPStrategy, LoadFromBootDiskFSDP)
-from demo.lightning.checkpoint.multinode.train import (DemoTransformer,
-                                                       init_processes)
 from google.cloud import storage
 from lightning import Trainer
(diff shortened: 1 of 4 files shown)


abhibyreddi:
I originally did not have `reader` and `writer` for the `FSSpecFSDPStrategy` in `__init__` but [the feedback I got at that time](https://github.com/GoogleCloudPlatform/dataflux-pytorch/pull/144#discussion_r1797487974) was that 1. it's generally a good practice to init them once and reuse 2. init-ing them in save and load methods would affect the times we report at the end.

abhibyreddi (approved):

awonak (approved):
LGTM. Nice improvement.

MattIrv:
> I originally did not have `reader` and `writer` for the `FSSpecFSDPStrategy` in `__init__` but [the feedback I got at that time](https://github.com/GoogleCloudPlatform/dataflux-pytorch/pull/144#discussion_r1797487974) was that 1. it's generally a good practice to init them once and reuse 2. init-ing them in save and load methods would affect the times we report at the end.

That feedback is valid in general, but since the reader/writer are tied to a specific path it introduces the limitation that you can't use different paths without completely recreating the strategy, which breaks some things. For example if you were to try to save to a different directory on each iteration of checkpoint save.

jdnurme (approved):

PR #175, opened 2024-11-04
Update auto wrap policy and remove duplicate load in trainer.fit

MattIrv:
This updates the benchmark to use the FSDP wrap policy defined in https://lightning.ai/docs/pytorch/stable/advanced/model_parallel/fsdp.html#identify-large-layers. 

Testing on a machine with 4 GPUs, before this change I get a CUDA OOM at a certain number of parameters for my model, and after the change I do not. 

I also removed the checkpoint restore path from the call to `trainer.fit` during load since this was making the trainer restore from that checkpoint an extra time, which is time consuming, and also explicitly deleting trainer/strategy/model prior to resetting them in case it helps prevent memory leaks.

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
@@ -5,15 +5,17 @@ import time
 
 import torch
 import torch.distributed
-from demo.lightning.checkpoint.multinode.strategies import (
-    DatafluxFSDPStrategy, FSSpecFSDPStrategy, LoadFromBootDiskFSDP)
-from demo.lightning.checkpoint.multinode.train import (DemoTransformer,
-                                                       init_processes)
+import torch.nn
 from google.cloud import storage
 from lightning import Trainer
 from lightning.pytorch.demos import WikiText2
 from lightning.pytorch.strategies import FSDPStrategy
 from torch.utils.data import DataLoader
+
+from demo.lightning.checkpoint.multinode.strategies import (
+    DatafluxFSDPStrategy, FSSpecFSDPStrategy, LoadFromBootDiskFSDP)
+from demo.lightning.checkpoint.multinode.train import (DemoTransformer,
+                                                       init_processes)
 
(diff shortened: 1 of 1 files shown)


abhibyreddi (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -180,7 +196,7 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
             devices=os.environ.get("NUM_DEVICES", 'auto'),
             num_nodes=num_nodes,
         )
-        trainer.fit(model, dataloader, ckpt_path=new_ckpt_dir_path)
+        trainer.fit(model, dataloader)
```
jdnurme, line 199: Was this ckpt_path param unnecessary since it's provided to the strategy? This is validated for every strategy?

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -53,6 +55,9 @@ def validate(args):
 
 def get_strategy(args, project, model, ckpt_dir_path):
     strategy = None
+    policy = {
```
jdnurme, line 58: Just to clarify, does this makes it so that FSDP only applies to the specified layers, and other layers are not managed by the FSDP strategy?

jdnurme (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
MattIrv: Yes, passing the checkpoint path here causes the trainer to look for an existing checkpoint to restore from, which isn't the behavior we want since we manually restore a checkpoint in the next line.

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
MattIrv: Right, other top-level layers in the model will be duplicated across all copies. Since these are the only large ones it performs as expected.

PR #168, opened 2024-10-30
PyTorch distributed checkpoint async save demo

awonak:
Add demo code for running PyTorch distributed async save checkpoints.

Also updates multinode checkpoint demo to specify the checkpoint save path.

Internal issue: [b/369834594](http://b/369834594)

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #176, opened 2024-11-04
Update save/load print statements for FSDP benchmark

MattIrv:
Now it prints each save/load time individually and also all at the end. This makes it easier to understand how the data differs from sample to sample.

### DIFF
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
@@ -201,8 +201,10 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
         avg_load_time = statistics.mean(load_checkpoint_times)
         avg_save_time = statistics.mean(save_checkpoint_times)
         print_times(args, avg_save_time, avg_load_time)
-        print(f"All save times: {save_checkpoint_times}")
-        print(f"All load times: {load_checkpoint_times}")
+        if not args.load_only:
+            print(f"All save times: {save_checkpoint_times}")
+        if not args.save_only:
+            print(f"All load times: {load_checkpoint_times}")
 
 
 if __name__ == "__main__":
--- a/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
+++ b/dataflux_pytorch/benchmark/checkpointing/multinode/train.py
@@ -5,15 +5,16 @@ import time
 
 import torch
 import torch.distributed
-from demo.lightning.checkpoint.multinode.strategies import (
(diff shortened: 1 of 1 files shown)


abhibyreddi (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -186,12 +192,17 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
...
+            print(
+                f"Loaded checkpoint from {new_ckpt_dir_path} in {end - start} seconds"
+            )
         load_checkpoint_times.append(end - start)
 
     if torch.distributed.get_rank() == 0:
         avg_load_time = statistics.mean(load_checkpoint_times)
+        avg_save_time = statistics.mean(save_checkpoint_times)
         print_times(args, avg_save_time, avg_load_time)
+        print(f"All save times: {save_checkpoint_times}")
```
jdnurme, line 204: nit: we could just put `if args.save_only:` around the print for loads to make output cleaner in that case.

jdnurme (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
```diff
@@ -142,14 +143,19 @@ def main(ckpt_dir_path: str, ckpt_restore_path: str = ""):
     )
     trainer.fit(model, dataloader)
     print(f"Saving checkpoint to {ckpt_dir_path} {num_save_calls} times.")
-    start = time.time()
+    save_checkpoint_times = []
```
awonak, line 146: Also consider using [torch.cuda.Event()](https://pytorch.org/docs/stable/generated/torch.cuda.Event.html) to `record()` and measure `elapsed_time()`. 

See an example usage in section 2.4 of https://pytorch.org/tutorials/intermediate/FSDP_tutorial.html#how-to-use-fsdp

Whatever you decide here, I'll update my PR #168 to match.

awonak (approved):

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
MattIrv: Our calls to save/load checkpoints here already enforce synchronization at the end in the underlying pytorch distributed logic, so I don't think I need to do anything special. Your case might be different if the nodes aren't already synchronized at that point.

dataflux_pytorch/benchmark/checkpointing/multinode/train.py
MattIrv: Done

PR #182, opened 2024-11-07
move demo model class to common lib

akansha1812:
Move duplicate code from single-node/model.py and distributed/model.py to a shared library.


demo/lightning/text_based/demo_model.py
```diff
@@ -0,0 +1,99 @@
...
+        self.encoder = encoder
+        self.decoder = decoder
+
+    def configure_optimizers(self):
+        """
+        This function sets the loss parameter (lr) which determines the rate at which the model "learns".
+        The parameters pass is a direct-passthrough of any existing parameters from the LightingModule.
+        """
+        optimizer = optim.Adam(self.parameters(), lr=1e-3)
+        return optimizer
```
awonak, line 58:
```suggestion
        return optim.Adam(self.parameters(), lr=1e-3)
```
nit: Since we're refactoring, let's remove the local variable assignment and just return the optimizer instance.

awonak (approved):

jdnurme (approved):

PR #184, opened 2024-11-11
Updating .gitignore

Yash9060:
Adding VS Code workspace to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,6 +153,9 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+# VSCode Workspace
+*.code-workspace
+
 # PyCharm
 #  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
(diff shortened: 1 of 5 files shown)


.vscode/settings.json
```diff
@@ -17,4 +17,10 @@
     },
     "files.insertFinalNewline": true,
     "files.trimFinalNewlines": true,
+    "yapf.interpreter": [
+        "./.venv/bin/python"
```
MattIrv, line 21: I'm not sure there's anything else in our setup that assumes your virtual environment is at `.venv` yet, is there? I think it's fine to do this since that's where it should be, but I wonder if there's anything else we can configure here that would do something like automatically make vscode set up the virtual environment there?

I also thought we had solved this problem before by activating a virtual environment within vs code, which this path would presumably conflict with. Can you double check?

MattIrv (approved):

.vscode/settings.json
jdnurme: I'm worried this could cause overwrites of people's existing configuration. We may not want to push this

Yash9060: > I'm not sure there's anything else in our setup that assumes your virtual environment is at .venv yet, is there? 

You are correct, there isnt anything which would point to this exact location. But as long as the we follow the documentation on developer setup (which is essentially just 
`Open Command palette -> Python: Create environment -> Keep on selecting defaults`. )

It would create venv in this exact location. But as an additional safety measure I can add a comment saying make sure this path is same as Python interpreter path https://screenshot.googleplex.com/9qi5SxbPk4TpamH 

> I also thought we had solved this problem before by activating a virtual environment within vs code, which this path would presumably conflict with. Can you double check? 

I already had active python venv within vscode and yet it didnt resolve the issue for me. Can you expand upon why we would have conflict since the python of interpreter and the one being passed to yapf & isort is same ?

awonak: Yash & I were chatting in another space and here's what we discovered:
1) The workspace venv is separate from the vs code extension venv. They are intentionally kept separate from each other. You don't want your workspace extensions to be dependencies of your work project.
2) VS Code extensions are installed in a different location from your project workspace venv: https://code.visualstudio.com/docs/editor/extension-marketplace#_where-are-extensions-installed

So this suggests there was a problem with the extension installation on that machine. 

My suggestion is to uninstall/reinstall the extension and try again without the changes in this PR.

jdnurme (approved):

PR #186, opened 2024-11-11
Remove image segmentation duplicates

akansha1812:
Reuse demo/image-segmentation functions in /demo/lightning/image-segmentation.
- unet3d.py
- losses.py
- layers.py is part of unet3d.py
- pytorch/get_train_transforms 
- pytorch/RandBalancedCrop

Tested complete run for /demo/lightning/image-segmentation/train.py. Tested imports for demo/image-segmentation/train.py.


awonak (approved):

setup.py
```diff
@@ -13,7 +13,7 @@
     packages=[
         "dataflux_core", "dataflux_pytorch", "dataflux_pytorch.lightning",
         "dataflux_pytorch.multipart_upload", "demo.lightning.text_based",
-        "demo.lightning.checkpoint.multinode"
+        "demo.lightning.checkpoint.multinode","demo.image_segmentation"
```
MattIrv, line 16: Maybe not for this PR but I'm not sure the demo code actually needs to be included here. Do things break if we remove it?

MattIrv (approved):

setup.py
MattIrv: Can we just use `-m`? In general we should be using `-m`

akansha1812: I think we can do that. Along with Kokoro we have to update some of the readme too. 
I'll create a new PR to remove demo from setup.py

abhibyreddi (approved):

setup.py
akansha1812: Resolving the comment for now.

PR #190, opened 2024-11-13
Update README with more async details

awonak:
Add clarification to main README on the different types of async checkpoint save support & demo code.


Yash9060 (approved):

README.md
```diff
@@ -178,9 +178,11 @@ Note that saving or restoring checkpoint files will stage the checkpoint file in
...
+PyTorch Lightning Singlenode - Our lightning checkpointing implementation has built-in support for the experimental [AsyncCheckpointIO](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.plugins.io.AsyncCheckpointIO.html#asynccheckpointio) featureset. This is an optimzation that allows for non-blocking `save_checkpoint` calls during a training loop. For more details on our support for this feature please see the [checkpoint README](https://github.com/GoogleCloudPlatform/dataflux-pytorch/tree/main/demo/lightning/checkpoint/singlenode#using-asynccheckpointio). This lightning feature is only supported for single-node executions.
 
-To allow for asyncronous saves with multinode executions we utilize PyTorch's [Async Distributed Save](https://pytorch.org/tutorials/recipes/distributed_async_checkpoint_recipe.html) to allow for similar non-blocking checkpoint operations. This implementation can be found [here](https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/demo/lightning/checkpoint/multinode/strategies.py#L99).
+PyTorch Singlenode - TODO
```
MattIrv, line 183: If this is going to be checked in now, let's remove this from the readme until it's ready

README.md
```diff
@@ -178,9 +178,11 @@ Note that saving or restoring checkpoint files will stage the checkpoint file in
...
+PyTorch Lightning Singlenode - Our lightning checkpointing implementation has built-in support for the experimental [AsyncCheckpointIO](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.plugins.io.AsyncCheckpointIO.html#asynccheckpointio) featureset. This is an optimzation that allows for non-blocking `save_checkpoint` calls during a training loop. For more details on our support for this feature please see the [checkpoint README](https://github.com/GoogleCloudPlatform/dataflux-pytorch/tree/main/demo/lightning/checkpoint/singlenode#using-asynccheckpointio). This lightning feature is only supported for single-node executions.
```
MattIrv, line 181: nit: Here and below, "single-node" and "multi-node" not "singlenode" or "multinode"

README.md
```diff
@@ -178,9 +178,11 @@ Note that saving or restoring checkpoint files will stage the checkpoint file in
...
 
-To allow for asyncronous saves with multinode executions we utilize PyTorch's [Async Distributed Save](https://pytorch.org/tutorials/recipes/distributed_async_checkpoint_recipe.html) to allow for similar non-blocking checkpoint operations. This implementation can be found [here](https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/demo/lightning/checkpoint/multinode/strategies.py#L99).
+PyTorch Singlenode - TODO
+
+PyTorch Multinode - To allow for asyncronous saves with multinode executions we utilize PyTorch's [Async Distributed Save](https://pytorch.org/tutorials/recipes/distributed_async_checkpoint_recipe.html) to allow for similar non-blocking checkpoint operations. This implementation can be found [here](https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/demo/lightning/checkpoint/multinode/strategies.py#L99).
```
MattIrv, line 185: I wouldn't recommend linking directly to the implementation here, but rather to whatever demo code we have that shows how to use it.

awonak: Currently we have an example usage in our benchmark folder path. Do you think it would be better to move this into our demo folder path, or is it okay as is?

MattIrv: It depends, is it really a benchmark or is it more of a demo? We could also link to the code library that we provide for this, it's just that the current link is to an internal implementation of a specific method and not the actual library code that users will use.

awonak: I have updated this paragraph to reference both the class for implementation and a reference to the `train_async_save.py` script for usage.

abhibyreddi (approved):

PR #198, opened 2024-11-15
LLAMA2 Simulated version

Yash9060:
Fixes b/378909790 
I have added a doc in the above issue which describes at how I calculated the correct size for each model. 

There is lot of code duplication between multinode/simulated demo/benchmarks. I will raise a follow up PR to move all the code into common directory and re-use it everywhere.

I will also update documentation on how to run benchmark code in follow up PR. 

Also removed certain piece of code from benchmark/simulated/benchmark.py which I accidentally checked in an earlier PR. 

- [X] Tests pass


dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
```diff
@@ -0,0 +1,119 @@
...
+        print(f"Time taken to save checkpoint:\
+                {statistics.mean(save_checkpoint_times):.4f} seconds (stdev {statistics.stdev(save_checkpoint_times):.4f})"
+              )
+        print(f"All save times: {save_checkpoint_times}")
+        print(f"Time taken to load checkpoint:\
+                 {statistics.mean(load_checkpoint_times):.4f} seconds (stdev {statistics.stdev(load_checkpoint_times):.4f})"
+              )
+        print(f"All load times: {load_checkpoint_times}")
+
+        tensor_size_per_instance = sum(v.element_size() * v.numel()
```
jdnurme, line 92: nit: Should probably call this "total_data_volume" or something indicating its summing all tensors, not just the size of a single tensor.

dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
```diff
@@ -0,0 +1,119 @@
...
+                                       for v in state_dict.values())
+        total_size_bytes = tensor_size_per_instance * world_size
+
+        print(f"Size of distributed tensors (rank {rank}):\
```
jdnurme, line 96: this shouldn't be gated behind `if rank == 0`

dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
```diff
@@ -0,0 +1,119 @@
+import os
+import socket
+import statistics
+import time
+
+import torch
+import torch.distributed as dist
+
+from demo.lightning.checkpoint.simulated.llama2 import (
```
jdnurme, line 9: This doesn't need to be addressed here, but we should probably restructure our imports such that benchmark doesn't import directly from "demo" or vice versa. Imports probably need to be structured into a "components" or "building_blocks" location outside of both.

demo/lightning/checkpoint/simulated/llama2.py
```diff
@@ -0,0 +1,289 @@
...
+    os.environ['MASTER_ADDR'] = MASTER_ADDR
+    os.environ['MASTER_PORT'] = MASTER_PORT
+    dist.init_process_group("gloo", rank=rank, world_size=world_size)
+
+
+def cleanup() -> None:
+    dist.destroy_process_group()
+
+
+def time_checkpoint_operation(benchmark_strategy: BenchmarkStrategy,
```
jdnurme, line 182: Out of scope for this CL, but in the future we can save a lot of duplicate code if we use a timer decorator function defined in a single location and used on each method with `@timer_function`

demo/lightning/checkpoint/simulated/llama2_utils.py
```diff
@@ -0,0 +1,161 @@
...
+        add_adamw_states(state_dict, prefix, 'ff.w2',
+                         (hidden_dim, intermediate_dim), empty)
+        add_adamw_states(state_dict, prefix, 'ff.w3',
+                         (intermediate_dim, hidden_dim), empty)
+
+        # FFN norm AdamW states
+        add_adamw_states(state_dict, prefix, 'ffn_norm', (hidden_dim, ), empty)
+
+
+def create_llama2_state_dict(world_size: int,
```
jdnurme, line 115: We may want to specify in a comment or in function name that this is generating distributed weights according to the FSDP style split, where each node does only gets a `1/world_size` volume of data.

demo/lightning/checkpoint/simulated/llama2.py
```diff
@@ -0,0 +1,289 @@
...
+        print(f"Time taken to load checkpoint:\
+                 {statistics.mean(load_checkpoint_times):.4f} seconds (stdev {statistics.stdev(load_checkpoint_times):.4f})"
+              )
+        print(f"All load times: {load_checkpoint_times}")
+
+        tensor_size_per_instance = sum(v.element_size() * v.numel()
+                                       for v in state_dict.values())
+        total_size_bytes = tensor_size_per_instance * world_size
+
+        print(f"Size of distributed tensors (rank {rank}):\
```
jdnurme, line 256: This probably shouldn't be behind `if rank == 0:`

demo/lightning/checkpoint/simulated/multiprocessing_train.py
```diff
@@ -220,6 +220,7 @@ def time_checkpoint_operation(benchmark_strategy: BenchmarkStrategy,
         if i % world_size == rank:
             template_state_dict[f'dummy_tensor_{i}'] = torch.empty(
                 tensor_size, 1000)
+
```
jdnurme, line 223: was this change meant to sneak in here?

jdnurme (approved):

dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
Yash9060: Ack

dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
Yash9060: Any specific reason for printing this from all the nodes ? I mean I understand that  every node prints it's own size information, which may not be exactly equal but the size is approximately similar and shouldn't be a big deal right ?

demo/lightning/checkpoint/simulated/llama2.py
Yash9060: Ack

demo/lightning/checkpoint/simulated/llama2.py
Yash9060: Same as above

demo/lightning/checkpoint/simulated/multiprocessing_train.py
Yash9060: Removed

demo/lightning/checkpoint/simulated/llama2_utils.py
Yash9060: Done

dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def main() -> None:
+    world_size = int(os.getenv("WORLD_SIZE"))
+    project = os.getenv("PROJECT")
+    ckpt_dir_path = os.getenv("CKPT_DIR_PATH")
+    sample_count = int(os.getenv("SAMPLE_COUNT", 8))
+    use_fsspec = os.getenv("USE_FSSPEC",
+                           "False").lower() in ("true", "1", "yes")
+    model_parameter_size = os.getenv("MODEL_PARAMATER_SIZE")
```
awonak, line 112: nit: Consider making the model parameter sizes a tuple defined in llama2_utils which can be used for validation here and also with args.choices for the demo. This will help validate input sooner and help with documentation.

awonak (approved):
Nice work, thanks for the additional documentation and links to articles for context!

dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
Yash9060: For now, I think I want to keep all the variables limited to single file, instead of splitting them across llama2 and llama2_utils

dataflux_pytorch/benchmark/checkpointing/simulated/llama2.py
Yash9060: Changed it to local_tensor_size

PR #200, opened 2024-11-20
Add demo script for single-node  async checkpointing

awonak:
Fixes b/369874146

Example demo outputs:

```
$ python demo/checkpointing/train.py --project=gcs-tess --bucket=awonak-us-west1 --embedding-dim=10 --num-epochs=3 
Epoch 1/3, Train Loss: 449945.0625
Epoch 2/3, Train Loss: 409629.6875
Epoch 3/3, Train Loss: 380471.9375
Average checkpoint save time: 0.8570 seconds (stdev 0.1997)
Total run time: 1m53s

$ python demo/checkpointing/train.py --project=gcs-tess --bucket=awonak-us-west1 --embedding-dim=10 --num-epochs=3 --use-async
Epoch 1/3, Train Loss: 456512.0000
Epoch 2/3, Train Loss: 413212.9375
Epoch 3/3, Train Loss: 382256.8125
Average checkpoint save time: 0.0009 seconds (stdev 0.0009)
Total run time: 1m36s
```

Note: I will update the main README with single node async details once either this or #190 gets merged. 

- [x] Tests pass


demo/checkpointing/train.py
```diff
@@ -0,0 +1,166 @@
...
+        model.train()
+        for inputs, labels in dataloader:
+            optimizer.zero_grad()
+            outputs = model(inputs)
+            labels = labels.type(torch.FloatTensor)
+            loss = criterion(outputs, labels)
+            loss.backward()
+            optimizer.step()
+
+        losses.append(loss.detach().numpy())
```
jdnurme, line 146: Why is losses being tracked here? It doesn't seem like we use this data.

demo/checkpointing/train.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Save checkpoint
+        save_time = time.time()
+        ckpt_path = f'single-node/async/checkpoint_{epoch}.pth'
+        ckpt.save_checkpoint(ckpt_path, model.state_dict(), args.use_async)
+        save_checkpoint_times.append(time.time() - save_time)
+
+    # Clean up and report measurements.
+    ckpt.teardown()
+    print(f'Average checkpoint save time: '
```
jdnurme, line 157: We may want to print something different when the `--use-async` flag is enabled given that this timing will not represent the "checkpoint save time" in that instance.

jdnurme (approved):

demo/checkpointing/train.py
awonak: Whoops! This was left over from previous work where I was collecting losses for printing visualization charts using matplotlib.

PR #202, opened 2024-12-19
feat: Add forwards-compatibility with google-cloud-storage v3.x

andrewsg:


MattIrv (approved):

jdnurme:
I've run compatibility tests against the existing google-cloud-storage client and the pre-release. Have confirmed both versions are compatible with no visible performance impacts.

jdnurme (approved):

PR #203, opened 2025-03-25
Latest pytorch version

akansha1812:
Fixes  https://github.com/GoogleCloudPlatform/gcs-connector-for-pytorch/security/dependabot/2
The vulnerability arises in torch  2.3.1 version. Changing it to the latest version. 
- [ ] Tests pass

### DIFF
--- a/dataflux_pytorch/benchmark/requirements.txt
+++ b/dataflux_pytorch/benchmark/requirements.txt
@@ -1,2 +1,2 @@
-torch==2.3.1
+torch
 gcsfs
--- a/demo/lightning/checkpoint/requirements.txt
+++ b/demo/lightning/checkpoint/requirements.txt
@@ -1,2 +1,2 @@
-torch==2.3.1
+torch
 gcsfs


abhibyreddi (approved):

MattIrv (approved):

jdnurme:
I could be misreading it, but it looks like the error message [here](https://github.com/GoogleCloudPlatform/gcs-connector-for-pytorch/security/dependabot/2) is stating that it needs to be greater than 2.3.1. Also, doesn't this lock the version into 2.3.1? Should we change this in the .toml file instead?

akansha1812:
> I could be misreading it, but it looks like the error message [here](https://github.com/GoogleCloudPlatform/gcs-connector-for-pytorch/security/dependabot/2) is stating that it needs to be greater than 2.3.1. Also, doesn't this lock the version into 2.3.1? Should we change this in the .toml file instead and run it with a >=?

abhibyreddi:
> I could be misreading it, but it looks like the error message [here](https://github.com/GoogleCloudPlatform/gcs-connector-for-pytorch/security/dependabot/2) is stating that it needs to be greater than 2.3.1. Also, doesn't this lock the version into 2.3.1? Should we change this in the .toml file instead and run it with a >=?

I see in the toml file we have `torch >= 2.0.1` which should be fine (?) We had to lock the pytorch version just for the demos/benchmarks to `2.3.1` because we ran into some issues with `2.4` ([PR 148](https://github.com/GoogleCloudPlatform/gcs-connector-for-pytorch/pull/148)). This PR removes that version lock, right?

akansha1812:
> I could be misreading it, but it looks like the error message [here](https://github.com/GoogleCloudPlatform/gcs-connector-for-pytorch/security/dependabot/2) is stating that it needs to be greater than 2.3.1. Also, doesn't this lock the version into 2.3.1? Should we change this in the .toml file instead and run it with a >=?

I removed the version specified in the requirements.txt should pick the latest version https://pip.pypa.io/en/stable/user_guide/#installing-packages.

